PR #404, opened 2019-06-04
Peer Dependency documentation

JamesIves:
Adds a note about installing peer dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ You can install the component library into your project by running the following
 yarn add @unitedincome/components
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,8 @@ You can install the component library into your project by running the following
 yarn add @unitedincome/components
 ```
 
+You will also need to install the [required peer dependencies](https://github.com/UnitedIncome/components/blob/master/package.json#L109) into your project. You can learn more about peer dependencies [here](https://nodejs.org/es/blog/npm/peer-dependencies/).
```
matthewheck: I might add one more sentence here for explanation. Like: "We utilize peer dependencies because we often require dependencies that already exist in our application."

JamesIves: Agreed - pushed up a change.

PR #418, opened 2019-06-05
[APP-2784] Removing text-left

JamesIves:
Removes the `text-left` class from the cabinet as it's no longer required now everything lives in a portal.

This simply removes the left aligned restriction on the cabinet labels.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-2784)

### DIFF
--- a/components/molecules/Cabinet/Cabinet.js
+++ b/components/molecules/Cabinet/Cabinet.js
@@ -170,7 +170,6 @@ class Cabinet extends Component {
 
     const containerClasses = classNames(
       {
-        'uic--text-left': true,
         'uic--cabinet-container': true,
       },
       className


components/molecules/Cabinet/Cabinet.js
```diff
@@ -170,7 +170,6 @@ class Cabinet extends Component {
 
     const containerClasses = classNames(
       {
-        'uic--text-left': true,
```
RohitRajendran: This isn't going to cause any issues with current other uses of the cabinet right?

JamesIves: Shouldn't do, this was primarily used to enforce the alignment within the cabinet. Now that it gets yanked into a portal it doesn't make sense to restrict this.

PR #426, opened 2019-06-06
DropDown Portal

JamesIves:
Creates a portal for the DropDown component, allowing it to appear as a child of an `overflow: hidden` element. This subsequently fixes the animation issue on the itemization screens. 

react-select docs: https://react-select.com/advanced#portaling

* Load up the DropDown on the itemization screen. It will now get portalized and be visible. 


components/atoms/DropDown/DropDown.js
```diff
@@ -240,6 +251,7 @@ class DropDown extends Component {
               isDisabled={disabled}
               isLoading={loading}
               aria-label={label}
+              menuPortalTarget={this.portal}
```
RohitRajendran: Can we add a comment about why this is using a portal?

PR #427, opened 2019-06-06
[APP-2960] - disabled continue btn

RohitRajendran:
### Description
Supports actions when a user tries to click on continue even though it is disabled because of incomplete fields.

On the third click of the disabled continue button, a message will show above the button saying "You must answer this question before hitting continue." and will scroll you up to the first invalid field.

On the fourth click and every other click after that, it will jiggle the message.

### Testing Instructions
1. Try clicking continue multiple times on a card with required or invalid fields

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-2960)


components/atoms/Input/Input.js
```diff
@@ -322,6 +321,7 @@ class Input extends Component {
     };
 
     const showInvalidity = !disabled ? !this.isValid() : false;
+    const isEmpty = (value && value.length < 1) || !value;
```
JamesIves: Is this first check required as value defaults to an empty string anyway?

components/atoms/DropDown/DropDown.js
```diff
@@ -220,38 +215,59 @@ class DropDown extends Component {
...
-              aria-label={label}
-              {...optionProps}
-            />
-          </div>
-          {description && this.state.isValid ? (
-            <div className="uic--description">{description}</div>
-          ) : (
-            <div className="uic--validation-error">
-              {this.state.validationMessage || 'Invalid'}
+      <CardShellContext.Consumer>
```
JamesIves: Same note about the comment applies here!

components/atoms/DropDown/DropDown.js
```diff
@@ -1,12 +1,12 @@
 /** @module DropDown */
-import React, {Component} from 'react';
-import PropTypes from 'prop-types';
 import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
```
JamesIves: I'm gonna take another stab at linting import ordering again near the EOD today. Consistency!

components/atoms/DropDown/DropDown.js
```diff
@@ -300,6 +316,8 @@ DropDown.propTypes = {
   loading: PropTypes.bool,
   /** Determines if this field has been submitted or not. */
   isSubmitted: PropTypes.bool,
+  /** Displays error state for incomplete required fields */
+  showRequiredError: PropTypes.bool,
```
JamesIves: Same goes for the DropDown. Can we also add a Storybook story which stages a DropDown in the CardShell so we can keep track of this.

components/molecules/CardShell/CardShell.js
```diff
@@ -299,16 +371,25 @@ class CardShell extends Component {
...
-                          light
-                          type="submit"
-                          variant="secondary"
+                        <div
+                          className="uic--card-submit-wrapper"
+                          onClick={
+                            (isInvalid || disabled) &&
+                            this.onDisabledContinueClick
+                          }
+                          role="presentation"
```
JamesIves: 👍

components/molecules/CardShell/CardShell.js
```diff
@@ -203,6 +217,48 @@ class CardShell extends Component {
...
+      this.setState({shakeError: !this.state.shakeError});
+    }
+  }
+
+  /**
+   * Brings the first error field into view
+   * @returns {undefined}
+   */
+  scrollToFirstErrorField() {
+    const elements = document.getElementsByClassName('uic--error');
```
JamesIves: I think it would be more clearer to do something like `const firstVisibleError = document.getElementsByClassName('uic--error')[0];` It reads cleared when you do `firstVisibleError.scrollIntoView` I think.

components/atoms/Input/Input.js
```diff
@@ -508,6 +522,8 @@ Input.propTypes = {
   className: PropTypes.string,
   /** Sanitizes the input when passed back by the onChange handler. */
   sanitize: PropTypes.bool,
+  /** Displays error state for incomplete required fields */
+  showRequiredError: PropTypes.bool,
```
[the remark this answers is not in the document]
RohitRajendran: No, we don't want it to be the default behavior since majority of our inputs are required.

components/atoms/DropDown/DropDown.js
RohitRajendran: In the CardShell or just with it set to true? With the prop, it doesn't actually need to be in the card

components/molecules/CardShell/CardShell.scss
```diff
@@ -82,6 +88,40 @@ $active-border-size: 4px;
...
 
+  .uic--warning-message {
+    &.uic--card-required-error-msg {
+      margin-top: 1.5rem;
+
+      &.uic--card-msg-shake {
+        animation: shake 0.82s cubic-bezier(0.36, 0.07, 0.19, 0.97) both;
+        transform: translate3d(0, 0, 0);
+
+        @keyframes shake {
```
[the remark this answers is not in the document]
RohitRajendran: How about a class that we can just apply instead?

components/molecules/CardShell/CardShell.js
RohitRajendran: This might not always work though, radio buttons and checkboxes can be marked as required but don't have an error state. That's why we added the message at the bottom so that the user will still know there's something wrong in those cases.

components/atoms/DropDown/DropDown.js
JamesIves: Either works, it would be nice to see the DropDown in a card so perhaps that might be the best?

components/atoms/Input/Input.js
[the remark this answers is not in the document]
JamesIves: Disregard, just realized that this doesn't do exactly what I thought it did!

components/molecules/CardShell/CardShell.scss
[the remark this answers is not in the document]
JamesIves: That would probably be better!

components/molecules/CardShell/CardShell.js
JamesIves: Makes sense, at any rate can we change the variable name to `errors` to make it more clear?

PR #444, opened 2019-06-10
Version Tagging

JamesIves:
I would like to synchronize npm versions with a release tag in Github.

I would also like to start tagging develop builds with an alpha or beta version number, but that will come in a separate PR. (I mostly want to make sure this works first)


.circleci/config.yml
```diff
@@ -128,7 +128,7 @@ jobs:
           command: |
             npm version patch -m "Bump version to %s [skip ci]"
             npm publish
-            git push --set-upstream origin $CIRCLE_BRANCH
+            git push --set-upstream origin $CIRCLE_BRANCH --tags
```
matthewheck: this will certainly push any created tags to github, but I believe you're missing the operative part of creating the version tag. to the best of my knowledgee, at this point, this pr will do nothing.

CalMlynarczyk: `npm version patch` creates the tag. https://docs.npmjs.com/cli/version

PR #452, opened 2019-06-11
[APP-2982] - Fix card validation for inputs that use a mask and are not required

RohitRajendran:
### Description
If an input inside a card is not required but has a value, card validation will now make 

### Testing Instructions
1. You can test this in the Real Estate Flow > Mortgage Payment Info Card > Date Your PMI Will End field. The card should validate if that field has no value or if it has a valid date

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-2982)


components/atoms/ExpandCollapse/ExpandCollapse.scss
```diff
@@ -100,7 +100,7 @@ $transition-time: 0.5s;
   }
 }
 
-.uic--mcgonagall-card-wrapper {
+.uic--mcgonagall-card-shell {
```
RohitRajendran: Fixes expand and collapse animation in a card

components/molecules/CardSummaries/FormSummary/FormSummary.scss
```diff
@@ -18,6 +18,10 @@
       background-color: $table-alternate;
       border-bottom: solid 1px $light-gray;
       border-top: solid 1px $light-gray;
+
+      &:last-child {
+        border-bottom: 0;
```
RohitRajendran: Card shell already had a border and this prevents summary from doubling it

PR #464, opened 2019-06-13
[APP-2917] Input and DropDown spacing

JamesIves:
Couple of changes here...

* Formatted all markdown files with Prettier (for some reason this didn't happen before?)
* Made spacing consistent between all dropdown and input fields in a card. Should be 1.5rems between (This matches the Bootstrap grid, Monique has approved the spacing)
* Added `(Optional)` to inputs which are not marked as disabled in their label.
* Staged a QuestionCard story which matches this comp: https://app.zeplin.io/project/589b5c319e603a741d558f10/screen/5c7e7201429b3a6274bb644a
* Fixed an issue with the Input and DropDown field's disabled styles. They were not getting the correct styles if they were disabled *with* a value, I've fixed it and added a story which shows this state.

Look at them in Storybook, everything should be happy days.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-2917)

### DIFF
--- a/components/atoms/Button/Button.md
+++ b/components/atoms/Button/Button.md
@@ -7,11 +7,11 @@ The Button component displays a HTML button.
 You can use this component using one of the following import patterns.
 
 ```javascript
-import Button from '@unitedincome/components/dist/Button'
+import Button from '@unitedincome/components/dist/Button';
 ```
 
 ```javascript
-import {Button} from '@unitedincome/components'
+import {Button} from '@unitedincome/components';
 ```
 
 ## Example 🚀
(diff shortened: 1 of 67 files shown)


RohitRajendran:
Can you update the lint command so that prettier checks md files as well?

JamesIves:
> Can you update the lint command so that prettier checks md files as well?



> Can you update the lint command so that prettier checks md files as well?

Yep, done

RohitRajendran:
You'll have to make a client update to remove the optional tag from current uses of the input like this one:
image

RohitRajendran:
And remove everywhere that already has optional in the label like on the deposit screen 
image

PR #467, opened 2019-06-13
[APP-2800] Adding Percy pull request check

JamesIves:
This change checks to see if visual regression tests are needed. With these changes, when CircleCI builds `develop` it will test to see if an active pull request is open against the `master` branch. If it is it will pass the check and run the tests, if it is not it will still greenlight the build, but not run the Percy checks.

With these changes we'll be able to test to make sure that we are only running Percy before we attempt a release, which means we'll no longer have a new version cut before we see the visual changes.

* We'll need to try this in CircleCI, I don't know if there's a way to test these locally. You can test the bash scripts locally that I've put into the CircleCI though, you'll need an access token first with public repo read access saved as an env variable.

The bash script in the Circle config file wraps the node command. The reason for this is because CircleCI handles exit statuses of 1 as a build failure. With the wrapper the yarn command can exit and cease to run, but it won't fail the build. Technically this workflow will run on every commit to develop, it just won't run the visual tests on every commit unless a pull request is open. 

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-2800)


.circleci/config.yml
```diff
@@ -63,7 +63,20 @@ jobs:
 
       - run:
           name: Run Visual Regression Tests
-          command: yarn snapshot
+          command: |
+            #!/usr/bin/env bash
+            # Sees if visual tests should be executed on the branch based on what pull requests are open.
+            # Executes the visual check command.
+            yarn testVisual
```
JamesIves: The `yarn` script gets wrapped in a bash script so the exit status doesn't bubble up to the Circle job. If this "fails" it means that a pull request isn't open.

.circleci/config.yml
```diff
@@ -140,7 +153,7 @@ workflows:
           filters:
             branches:
               only:
-                - master
+                - develop
```
JamesIves: Visual checks should only run on develop now with these changes.

PR #492, opened 2019-06-19
[APP-2988] Input Explanation Text

JamesIves:
Adds the `explanation` prop on top of the input field. I've had to make some changes to how the percent field works, but all of them I feel are probably for the best.

hhzuzuzuz

* Pull down the branch and go into Storybook.
* Find an input field and focus it.
* You'll see the explanation prop in action.
* Also added it to the simple form example within QuestionCard.


RohitRajendran:
What Nick said

labichn:
I'm just so disappointed.

RohitRajendran:
We expect so much better from you

matthewheck:
this is shite

JamesIves:
u lot are so mean

blakeduncan14:
Agree with the code coverage comments but otherwise lgtm

PR #515, opened 2019-06-26
[APP-2931] Box Component

JamesIves:
Adds the `<Box />` component, which lets you display a label, description and icon in a convenient box. 

When an icon is provided, the `illuminate` prop is linked to the parent components `selected` prop, which means when the box is in a selected state the illustration will activate.

image

image

* Load up the story in storybook and take a look.

This pull request does not include the BoxSelection component, that will fast follow. I wanted to seperate them out as they are separate tickets.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-2931)


components/atoms/Box/Box.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+    className
+  );
+
+  let boxIcon;
+
+  if (icon) {
+    /* The icon component is cloned here so we can enforce certain styles such
+      as height, width, and the selected state. */
+    boxIcon = React.cloneElement(icon, {
```
RohitRajendran: Does it really have to be cloned? Can’t you just do <icon illuminate={selected} className=“uic--box-illustration”/>

JamesIves: I'm cloning it to keep the props from the icon intact for height/width and color - this is mostly because of the size variation issue we talked about. The other option would be to add an `iconHeight` and `iconWidth` and `iconColor` prop. Either way works for me.

RohitRajendran: Shouldn’t those be set inside the Box component so that the icons are used consistently across all the boxes?

JamesIves: No because the viewbox of the svg's differs slightly, which means `10rem` on the car illustration is different from `10rem` on the ramen illustration.

JamesIves: Going ahead with changing the pattern for now and enforced the height/width on all illustrations, we'll need to make sure this gets fixed before it's production ready though. Monique will take a look next week.

components/atoms/Box/Box.js
```diff
@@ -0,0 +1,55 @@
...
+  );
+};
+
+Box.propTypes = {
+  /* Additional class names to apply to the container. */
+  className: PropTypes.string,
+  /** Additional style properties to apply to the container. */
+  style: PropTypes.objectOf(PropTypes.string),
+  /** The label to appear on the box. */
+  label: PropTypes.string,
```
RohitRajendran: This would probably be better named as header or title and should be required

PR #517, opened 2019-06-26
[APP-2933] BoxSelectCard

JamesIves:
Implements the `BoxSelectCard` which lets you make a box selection within a question card.

gofff22222

* Load up the card component in storybook.
* Test it in a flow.

Also adds a focus state to the Box component. To handle this I've had to make it stateful.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-2933)


components/organisms-complex/BoxSelectCard/BoxSelectCard.story.js
```diff
@@ -0,0 +1,115 @@
...
+stories.add('active', () => (
+  <BoxSelectCard {...defaultProps(false, false, true)} />
+));
+
+stories.add('editing and clears future', () => (
+  <BoxSelectCard {...defaultProps(false, true, false)} />
+));
+
+stories.add('collapsed', () => (
+  <BoxSelectCard {...defaultProps(true, false, false)} answers={['Yup']} />
```
RohitRajendran: We probably won't ever see the collapsed state but what would it be if we could? This doesn't seem right

components/molecules/BoxSelect/BoxSelect.js
```diff
@@ -0,0 +1,88 @@
...
+        'uic--h-100': true,
+      },
+      className
+    );
+
+    return (
+      <div className={containerClasses} style={style}>
+        {options.map((item) => {
+          return (
+            <div key={item.value} className="uic--col- uic--col-sm-4">
```
RohitRajendran: Incomplete class

components/molecules/BoxSelect/BoxSelect.js
```diff
@@ -0,0 +1,88 @@
...
+  className: PropTypes.string,
+  /** Additional style properties to apply to the container. */
+  style: PropTypes.objectOf(PropTypes.string),
+  /** The select options to appear within the box. */
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      /** The value of the select button, for example `yes` or `no`. */
+      value: PropTypes.string.isRequired,
+      /** The label to appear on the box. */
+      label: PropTypes.string,
```
RohitRajendran: Should be title

components/molecules/BoxSelect/BoxSelect.md
```diff
@@ -0,0 +1,48 @@
...
+
+```javascript
+<BoxSelect
+  name="boxGroup"
+  onChange={(name, value) => this.setState({[name]: value})}
+  value={this.state.boxGroup}
+  options={[
+    {
+      value: 'essentials',
+      label: 'Essentials',
```
RohitRajendran: Should all be title

components/organisms-complex/BoxSelectCard/BoxSelectCard.js
```diff
@@ -0,0 +1,137 @@
...
+BoxSelectCard.propTypes = {
+  /** Values to pass into  BoxSelect component. */
+  config: PropTypes.shape({
+    /** The select options to appear within the box. */
+    options: PropTypes.arrayOf(
+      PropTypes.shape({
+        /** The value of the select button, for example `yes` or `no`. */
+        value: PropTypes.string.isRequired,
+        /** The label to appear on the box. */
+        label: PropTypes.string,
```
RohitRajendran: This should be title

components/organisms-complex/BoxSelectCard/BoxSelectCard.story.js
```diff
@@ -0,0 +1,115 @@
...
+  editCard: () => true,
+  config: {
+    name: 'boxSelectQuestion',
+    value: store.get('boxSelectQuestion'),
+    options: object('config.options', [
+      {
+        value: 'essentials',
+        label: 'Essentials',
+        description:
+          'These are the basics: paying the bills and putting food on the table. They include your non-discretionary spending.',
```
RohitRajendran: If text for one of the rows is longer than the other, the sizes become inconsistent.

RohitRajendran: Screen Shot 2019-06-27 at 3 29 17 PM

components/molecules/BoxSelect/BoxSelect.story.js
```diff
@@ -44,7 +44,7 @@ const items = [
     value: 'healthcare',
     label: 'Healthcare',
     description:
-      'This spending keeps you healthy. It includes your out-of-pocket health expenses and private insurance costs.',
+      'This spending keeps f-pocket health expenses and private insurance costsf-pocket health expenses and private insurance costsf-pocket health expenses and private insurance costsyou healthy. It includes your out-of-pocket health expenses and private insurance costs.',
```
RohitRajendran: There’s a typo here btw

components/organisms-complex/BoxSelectCard/BoxSelectCard.md
```diff
@@ -0,0 +1,21 @@
...
+```
+
+```javascript
+import {BoxSelectCard} from '@unitedincome/components';
+```
+
+## Example 🚀
+
+```javascript
+<BoxSelectCard />
```
RohitRajendran: Better example

components/molecules/BoxSelect/BoxSelect.story.js
matthewheck: f-pocket definitely sounds like some weird insult from James.

components/atoms/Box/Box.js
```diff
@@ -1,55 +1,109 @@
...
+        'uic--box-highlighted': selected,
+        'uic--box-hover': this.state.hover,
+      },
+      className
+    );
+
+    let BoxIcon;
+
+    if (icon) {
+      BoxIcon = icon;
```
matthewheck: how about a ternary instead?

```
const BoxIcon = icon ? icon : null;
```

components/atoms/Box/Box.js
```diff
@@ -1,55 +1,109 @@
...
+        onMouseEnter={this.handleHover}
+        onMouseLeave={this.handleHover}
+      >
+        {BoxIcon && (
+          <BoxIcon
+            className="uic--box-illustration"
+            illuminate={selected || this.state.hover}
+            height="10rem"
+            width="10rem"
+            color
```
matthewheck: is `color` just a boolean? if so, would it be better to say something like `isColor` to make that clear?

components/atoms/Box/Box.spec.js
```diff
@@ -19,3 +20,44 @@ test('Box - renders with an icon', (t) => {
...
+    onClick: stub(),
+  };
+  const component = mount(<Box {...props} />);
+
+  component
+    .find('.uic--box')
+    .at(0)
+    .prop('onClick')({
+    stopPropagation: stub(),
+  });
```
matthewheck: this is funky indentation

components/atoms/Box/Box.spec.js
```diff
@@ -19,3 +20,44 @@ test('Box - renders with an icon', (t) => {
...
+    value: 'montezuma',
+    label: 'the best cat',
+    description: 'v ginger and v good',
+    onClick: stub(),
+  };
+  const component = mount(<Box {...props} />);
+
+  component.instance().handleHover();
+
+  t.deepEquals(component.state().hover, true, 'Hover state should be true.');
```
matthewheck: For this to be a particular powerful test, it'd probably be good to prove that `hover` had a value of `false` before handleHover is called.

components/molecules/BoxSelect/BoxSelect.js
```diff
@@ -0,0 +1,117 @@
...
+        'uic--row': true,
+        'uic--h-100': true,
+      },
+      className
+    );
+
+    return (
+      <div className={containerClasses} style={style}>
+        {options.map((item) => {
+          return (
```
matthewheck: shouldn't need a return here, I believe.

RohitRajendran: You do since it’s wrapped inside of {}

RohitRajendran: But you could also just rid of {} and the return

components/molecules/BoxSelect/BoxSelect.js
```diff
@@ -0,0 +1,117 @@
...
+
+    // Passes back the currently selected item to the parent.
+    if (onChange && name) {
+      onChange(name, value);
+    }
+  }
+
+  /** @inheritdoc */
+  componentDidMount() {
+    const components = document.getElementsByClassName('uic--box');
```
matthewheck: I don't love this approach. How about using [ref forwarding](https://reactjs.org/docs/forwarding-refs.html)? It looks like it's (more or less) attempting to solve this problem.

JamesIves: This gets the height of all of the elements in the options array after they have rendered, I don't think applying a ref for each element would be appropriate here?

components/atoms/Box/Box.spec.js
JamesIves: The default state of hover is false so it feels redundant to set it here

matthewheck: Until somebody accidentally changes it and this always passes because nothing is happening!

components/molecules/BoxSelect/BoxSelect.js
matthewheck: Yeah, I'm not sure there's a better way than this.

PR #532, opened 2019-07-02
activeLink

JamesIves:
Makes it so the route highlighting is dependent on the prop instead of internal state. This means we can pass down the current route, ie `/fish` to the navbar component, and then if the `activeLink` prop matches the nav items url it will highlight it on the bar.


components/molecules/Navbar/Navbar.js
```diff
@@ -92,8 +87,8 @@ class Navbar extends PureComponent {
       const listItemClasses = classNames({
         'uic--position-relative': true,
         'uic--d-inline-flex': true,
-        'uic--navbar__navigation-active':
-          this.state.active === `${type}-${index}`,
+        'uic--justify-content-center': true,
+        'uic--navbar__navigation-active': this.state.active === link,
```
RohitRajendran: A library article would not activate the library link

components/molecules/Navbar/Navbar.js
```diff
@@ -259,6 +254,8 @@ Navbar.propTypes = {
   staticLogo: PropTypes.node,
   /** Determines at what point the navigation bar should switch to the fixed state. Measured in pixels from the top. */
   transitionToFixed: PropTypes.number,
+  /** The current route which is provided to apply proper active highlighting on load. */
+  activeLink: PropTypes.string,
```
RohitRajendran: .isRequired

PR #540, opened 2019-07-03
[APP-3032] Social Media Icons

JamesIves:
Adds `FacebookIcon`, `TwitterIcon` and `LinkedInIcon` to the component library.

Check them out in Storybook. Not much to test here.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-3032)


components/atoms/icons/FacebookIcon/FacebookIcon.js
```diff
@@ -0,0 +1,58 @@
...
+          </g>
+        </g>
+      </g>
+    </g>
+  </svg>
+);
+
+FacebookIcon.propTypes = {
+  /** The width of the with unit sizing (px, rem, etc). */
+  width: PropTypes.string,
```
RohitRajendran: What about default width and height?

components/atoms/icons/LinkedInIcon/LinkedInIcon.js
```diff
@@ -0,0 +1,62 @@
...
+      height,
+      ...style,
+    }}
+    viewBox="0 0 18 15"
+    version="1.1"
+    xmlns="http://www.w3.org/2000/svg"
+    xmlnsXlink="http://www.w3.org/1999/xlink"
+  >
+    <g
+      id="Symbols"
```
matthewheck: let's remove the id for this, "footer" and "social" since they're repeated in all icons.

components/atoms/icons/FacebookIcon/FacebookIcon.js
JamesIves: Should be included now

PR #548, opened 2019-07-08
Navbar Animation Issue

JamesIves:
Fixes an issue with the animation preventing it from things being clickable when the mobile navbar was present on the page.


JamesIves:
Going with a better approach

PR #551, opened 2019-07-08
Check for browser environment variables before using them

CalMlynarczyk:
### Description
We are still running into build errors trying to reference browser environment variables and APIs without checking for their existence first. I've added checks for all uses of `window`, `document`, and `navigator` that I found.


RohitRajendran:
It still seems weird to me that we’re having to add these checks since some of these components won’t work without document or window

CalMlynarczyk:
Test coverage resolved. Moved check into a shared function. Added a note to the README about global variables.

PR #561, opened 2019-07-10
[APP-3031] Create bubble navigation component

CalMlynarczyk:
### Description
Migrate the `FourthLevelNav` component and its used dropdown component (with modifications to match the dotcom-v2 comps) into this library. Add the ability to give each item its own custom color.

### Testing Instructions
Open the components in StoryBook.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-3031)


components/atoms/BubbleSelector/BubbleSelector.scss
```diff
@@ -0,0 +1,49 @@
...
+  border-radius: 2.4rem;
+  color: $slate;
+  cursor: pointer;
+  font-size: 1.4rem;
+  margin: 0 0.7rem;
+  max-width: 19.4rem;
+  padding: 0.7rem;
+  text-align: center;
+  text-transform: capitalize;
+  width: 100%;
```
RohitRajendran: This is making it always take up the full max-width instead of just the appropriate width for the content like in the comp. We should get rid of this and increase the padding to match the comps

components/atoms/BubbleSelector/BubbleSelector.scss
```diff
@@ -0,0 +1,49 @@
+.uic--bubble-selector__item {
```
RohitRajendran: There’s no font included

components/atoms/BubbleDropdown/BubbleDropdown.scss
```diff
@@ -0,0 +1,81 @@
+// Wrapper class gives specificity to override react-select styles
+.uic--bubble-dropdown {
```
RohitRajendran: There’s no font declared here too

components/atoms/BubbleDropdown/BubbleDropdown.scss
```diff
@@ -0,0 +1,81 @@
...
+    font-weight: 400;
+    padding: 1.5rem 0;
+    text-align: center;
+
+    &:first-child {
+      border-top: 0;
+    }
+  }
+
+  .uic--bubble-dropdown__option--is-focused {
```
RohitRajendran: There’s a blue background color being applied as you click on a dropdown option

components/molecules/BubbleNav/BubbleNav.js
```diff
@@ -0,0 +1,31 @@
...
+      <BubbleDropdown options={options} value={value} onChange={onChange} />
+    </div>
+  </Fragment>
+);
+
+BubbleNav.propTypes = {
+  /** Collection of option labels and optional colors */
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      value: PropTypes.string.isRequired,
```
RohitRajendran: Can you add descriptions for each of these as well

components/atoms/BubbleSelector/BubbleSelector.js
```diff
@@ -0,0 +1,65 @@
...
+        );
+      })}
+  </Fragment>
+);
+
+BubbleSelector.propTypes = {
+  /** Collection of option labels and optional colors */
+  options: PropTypes.arrayOf(
+    PropTypes.shape({
+      value: PropTypes.string.isRequired,
```
RohitRajendran: Descriptions for these props too

components/atoms/BubbleDropdown/BubbleDropdown.js
```diff
@@ -0,0 +1,94 @@
...
+        backgroundColor: color,
+      }),
+      menu: (provided) => ({
+        ...provided,
+        backgroundColor: color,
+      }),
+    };
+
+    return (
+      <Select
```
RohitRajendran: With the Select, is there an easy way to animate the dropdown opening to keep it consistent with the dropdown on the mobile version of this page https://dotcom-dev.unitedincome.team/library/the-retirement-solution-hiding-in-plain-sight

Definitely not a priority at this time

RohitRajendran:
Styling for the dropdown doesn’t look quite the same as the version in client. Seems to be missing some padding and the arrow is right next to the text instead of all the way to the right like so:
image

components/atoms/BubbleSelector/BubbleSelector.scss
CalMlynarczyk: Are you referring to `font-family`? Why do I need that?

RohitRajendran: To make sure it’s using the expected font family

CalMlynarczyk: Ah, I didn't notice that the browser button styles were overriding the font here.

components/atoms/BubbleDropdown/BubbleDropdown.scss
CalMlynarczyk: `react-select` is all `div`-based, so we don't have to worry about overriding the font family. It inherits from the page font family.

components/atoms/BubbleDropdown/BubbleDropdown.scss
CalMlynarczyk: Didn't catch that happening in Chrome.

components/atoms/BubbleDropdown/BubbleDropdown.js
```diff
@@ -0,0 +1,94 @@
...
+
+// react-select dropdown icon
+const DropdownIndicator = (props) => (
+  <components.DropdownIndicator {...props}>
+    <ArrowIcon height="1rem" fill={colors.white} />
+  </components.DropdownIndicator>
+);
+
+/** BubbleDropdown Component */
+export default class BubbleDropdown extends React.Component {
```
JamesIves: For consistency can we export after the class definition?

components/atoms/BubbleSelector/BubbleSelector.scss
CalMlynarczyk: Monique asked for two separate modes: one with uniform width and one with dynamic widths.

components/atoms/BubbleDropdown/BubbleDropdown.js
CalMlynarczyk: I added an animation to the icon to match, but the menu itself was not feasible at this time.

CalMlynarczyk:
@RohitRajendran The comps for the dotcom dropdown variation of the menu were different that what we had in the application right now, which is what I designed it to. However, Monique asked that I revert it to match the application so it looks similar now sans the placeholder text change when the menu is open.

RohitRajendran:
You can confirm with Monique but I believe the text is supposed to be at the center of the dropdown, right now it’s off center.

Screen Shot 2019-07-11 at 4 43 12 PM

CalMlynarczyk:
@RohitRajendran There was a slight offset that I fixed.

PR #570, opened 2019-07-12
CSS Changes

JamesIves:
Couple of small CSS changes

- Removes the max-width from the bubble selector as it causes text to wrap onto a second line when there's not enough space.
- Adds `line-height: 1.2` to the comorat fonts as specced out in Zeplin. For the larger font size adds `line-height: 1.25`.


CalMlynarczyk:
@JamesIves I changed the fixed width mode to take a value for the width instead of a boolean. This should rectify the issue with fixed width mode and better support what you want.

JamesIves:
Looks good to me!

PR #579, opened 2019-07-18
Navbar Jank Issue

JamesIves:
Reduces scroll jank by giving the navbar consistent spacing when it transitions to a fixed position, instead of eliminating 50px from the top of the page.


CalMlynarczyk:
Can you explain what exactly it is you're seeing and trying to fix? I don't know what you mean by "jank." I'm seeing some undesirable bouncing on the homepage in Chrome when you switch from fixed to unfixed mode, but I'm not sure if that's what you're referring to.

JamesIves:
> Can you explain what exactly it is you're seeing and trying to fix? I don't know what you mean by "jank." I'm seeing some undesirable bouncing on the homepage in Chrome when you switch from fixed to unfixed mode, but I'm not sure if that's what you're referring to.


That's the issue I'm fixing here, it's causing issues with background-attachment.

PR #594, opened 2019-07-19
Add optional group key to form summary

labichn:
### Description
The form summary component uses the group names as the React keys (see [here](https://github.com/UnitedIncome/components/blob/1d99ad2afdecd07e3a0e12217f12250a53df15a1/components/molecules/CardSummaries/FormSummary/FormSummary.js#L29))--which causes warnings if you have duplicate group names:
form-summary
warning

This PR allows answer groups to specify an optional key to use as the React key.

### Testing Instructions
Render a form summary with two answer groups that have the same name--there will be a React warning. Add a unique key name to one or both of the answer groups and re-render--there will not be a React warning.

### DIFF
--- a/components/molecules/CardSummaries/FormSummary/FormSummary.js
+++ b/components/molecules/CardSummaries/FormSummary/FormSummary.js
@@ -26,7 +26,7 @@ const FormSummary = ({shortTitle, editCard, answerGroups, to}) => {
         </div>
       </div>
       {answerGroups.map((group, index) => {
-        const key = group.groupName || `${shortTitle}-${index}`;
+        const key = group.key || group.groupName || `${shortTitle}-${index}`;
 
         const groupClass = classNames({
           'uic--card-summary': true,
--- a/components/molecules/CardSummaries/FormSummary/FormSummary.js
+++ b/components/molecules/CardSummaries/FormSummary/FormSummary.js
@@ -65,6 +65,8 @@ FormSummary.propTypes = {
   /** An array of objects containing the groups of answers to display. The object is broken down below. */
   answerGroups: PropTypes.arrayOf(
     PropTypes.shape({
+      /** The React key to use for the group of data */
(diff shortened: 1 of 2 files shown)


components/molecules/CardSummaries/FormSummary/FormSummary.js
```diff
@@ -26,7 +26,7 @@ const FormSummary = ({shortTitle, editCard, answerGroups, to}) => {
         </div>
       </div>
       {answerGroups.map((group, index) => {
-        const key = group.groupName || `${shortTitle}-${index}`;
+        const key = group.key || group.groupName || `${shortTitle}-${index}`;
```
dwolfand: do we care if the key is falsy like a `0`? Should we wrap in `nullOrUndefined()`?

matthewheck: Can you add `key` to the proptypes below?

Also, should this be optional or mandatory?

labichn: It should be optional.

matthewheck: I can see some value to requiring unique keys throughout as opposed to only adding them as a fallback when you see an error (more proactive instead of reactive), but it is definitely more onerous.

labichn: I'm assuming the form summaries to-date have unique answer groups, or this probably would have come up before. I don't want to force changes to existing code, but I understand your point.

labichn: Keys have to be strings, and I think there's value to interpreting the empty string as falsy in this context.

labichn: Documented in 24e65fa.

PR #623, opened 2019-07-30
[APP-3020] [APP-3181] SVG Scoping and Optimization

JamesIves:
This pull request fixes a number of issues we've been having with all of our new SVG illustrations, and as a result fixes a massive bug we've been having in the marketing site where our illustrations were not rendering correctly in IE and Safari.

* Correctly scopes all of the svgs using an id system which makes sense. Instead of using the defaulted export name, instead it now follows `#component-name-1` anytime a clip or fill rule is required. This prevents name clashing and style bleeding between svgs when multiple are used on the same page.
* Re-imported all of the svgs from the first round so their viewbox is now `0 0 140 140`, this should close `APP-3020` which has been open for some time.
* Ran all of the svgs through imageoptim, which has reduced the payload for each svg anywhere between 5-20%.
* Removed inline `<style>` tags in favor for the appropriate svg prop, ie `fill="#fff"`.
* Fixed a typo in the compass story.
* Fixed an issue where the health meter illuminated story wasn't rendering correctly. 

Screen Shot 2019-07-29 at 10 38 57 PM

* Look at each illustration in storybook, they should all look the same as they did before. They should also be the same size which was a bug before (Good way to test this is by looking at the car illustration, and then looking at the lamp, previously they were different sizes).
* Test each illustration in Safari and Firefox also, as these browsers tend to have different opinions on how svgs should render.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-3020)
- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-3181)


RohitRajendran:
Is having id's going to cause issues if we have multiple of the same svg on the same page?

JamesIves:
No, because they are correctly scoped. They require the ids for the gradient pathing, as far as I know there’s no other way to do this with an svg. Plus these types of illustrations usually only appear once per page anyway.

dwolfand:
Yeah, I like that idea of passing in a prod to make them unique if required - any reason not to make that change?

JamesIves:
Not that I can think of, I’ll just default them to the name of the component.

dwolfand:
Also, what is this imageoptim voodoo magic?! Its really impressive

JamesIves:
It’s an app for Mac! It’s been around forever

https://imageoptim.com/api

There’s other svg and other optimizers out there that can achieve the same thing too.

PR #641, opened 2019-08-06
Unstyled Button

JamesIves:
Utility class for unstyled buttons.


components/utilities/Styles/buttons.md
```diff
@@ -0,0 +1,17 @@
...
+These classes get imported as a part of the util stylesheet.
+
+```scss
+@import '@unitedincome/components/dist/util.css';
+```
+
+## Example 🚀
+
+```javascript
+<button className="uic--button-unstyled" onClick={() => console.log('hello')}>Hello</button>
```
RohitRajendran: Can you also explain when and why you'd use this?

JamesIves: Added an explanation in the buttons.scss file and the story.

matthewheck: Is there anything you can point to in the client or marketing site that uses this? I understand in theory what the goal is here, but I'm not sure when it's used in practice.

PR #644, opened 2019-08-09
[Innovation Sprint] Accessibility Changes for Components

JamesIves:
* Removes `outline: 0` and `outline: none`. These properties now get toggled off using the `:focus-visible` polyfill. In the Storybook preview you can click on a component, but you will not see the focus state. If you click tab on your keyboard you will. Magic! Learn more: https://a11yproject.com/posts/never-remove-css-outlines/

* Fixes some general rules that were flagged.

* Added the `:focus-visible` polyfill to the preview and added a note in the readme file.

Click on the button, you'll see no outline. Tab to the button, you'll see an outline.

Innovation sprint!


components/molecules/Modal/Modal.js
```diff
@@ -23,10 +23,7 @@ export const ModalInterior = ({
   });
 
   return (
-    <aside
-      className="uic--h-100 uic--w-100 uic--d-flex uic--align-items-center uic--justify-content-center"
-      role="dialog"
```
RohitRajendran: How come we got rid of the role?

matthewheck: seems like it should be here. https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html

JamesIves: The accessibility validator in the component library says it shouldn't - from what I've read `aside` is also a valid choice for modal components.

PR #667, opened 2019-08-19
[APP-3225] Mobile Navbar State

JamesIves:
Changes the navbar mobile state to match the new comps.

Also allows for a `staticLogoMobile` prop if a different SVG needs to render on those screen sizes.


components/molecules/Navbar/Navbar.js
```diff
@@ -230,10 +243,20 @@ class Navbar extends PureComponent {
         <div className="uic--navbar__logo uic--d-flex uic--align-items-center">
           {logoLink ? (
             <LinkWrapper {...linkProps}>
-              {fixed ? fixedLogo : staticLogo}
+              {fixed
```
RohitRajendran: Should just make this a const so that you don't have to repeat it

JamesIves: Yeah good idea

PR #702, opened 2019-09-06
Add optional debouncing to Input's onChange handler

CalMlynarczyk:
### Description
Add optional debouncing to the `onChange` handler for the `atoms/Input` component. The `onChange` option inside of this prop will fire using debouncing to prevent the handler from running too often. The `onChange` prop on the `Input` component will _not_ fire using debouncing. Separating these two allows us to create separate event handlers for immediate and delayed invocation; for example, to fire off a cheap handler like writing to the state store on every change, but to send the input value in an expensive AJAX request only once the user stops typing.

The optional `debounce` prop matches the options available in [`react-delay-input`](https://www.npmjs.com/package/react-delay-input). I originally tried to implement this myself but found I was going to have to write a lot of the same things this library did in order to manage debouncing event handlers within the React component lifecycle.

### Testing Instructions
1. Open the _debounce_ story for `atoms/Input`
2. Type in the input field
3. You should see the immediate output change as you type, but the debounce output should change only once you stop typing
4. You can use the `debounce` option under _Knobs_ to change the timing and behavior of the debouncing functionality


CalMlynarczyk:
@RohitRajendran had a better idea for how to implement the `onChange` handler for the McGonagal flow that this change supported.

PR #705, opened 2019-09-11
CardContentDescription

JamesIves:
Adds the `CardContentDescription` component. Sits within a card, and displays content whenever the conditions are met. The handler is debounced so it will only fetch once every 1000ms.

Docs are in the README.

* Load the `CardContentDescription` story, and type in the box. If the string is greater than 3 it will display the message. It won't fire while the user is typing.


components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    return (
+      <div className={containerClasses} style={style}>
+        <hr />
+        <div className="uic--mcg-subhead-2-text">
+          {isFetching ? (
+            <Spinner height="2rem" width="2rem" fill={colors.midnight} />
+          ) : (
+            data
```
RohitRajendran: Couldn't we just use children here instead?

components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -0,0 +1,105 @@
...
+      className
+    );
+
+    if ((!isFetching && !data) || !valid) {
+      return null;
+    }
+
+    return (
+      <div className={containerClasses} style={style}>
+        <hr />
```
RohitRajendran: This top HR is going to cause issues on the credit card flow where there's already a line from the expandcollapse that's used

JamesIves: You can now toggle `removeTopBorder` and `removeBottomBorder` to `false` to hide each one.

components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -0,0 +1,115 @@
...
+    if (values !== prevValues && valid) {
+      this.setState({
+        isFetching: true,
+        valid,
+      });
+
+      this.fetchData();
+    } else if (!valid && valid !== this.state.valid) {
+      // If the field is invalid then the following state causes the text to hide.
+      this.setState({
```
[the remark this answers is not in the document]
JamesIves: Already have, it's meant to be hidden

components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -0,0 +1,115 @@
...
+  /* Additional class names to apply to the container. */
+  className: PropTypes.string,
+  /** Additional style properties to apply to the container. */
+  style: PropTypes.objectOf(PropTypes.string),
+  /** HTML element that should appear when the component is not fetching. */
+  children: PropTypes.node.isRequired,
+  /** An object of values which the component will keep track of to see if there's a change. */
+  values: PropTypes.objectOf(PropTypes.string).isRequired,
+  /** The function which will fire when the validation function passes, and if a value has changed. */
+  onChange: PropTypes.func.isRequired,
```
RohitRajendran: onChange, values, and validate would not be required if it's just being used to display static text

components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    return (
+      <div className={containerClasses} style={style}>
+        {!removeTopBorder && <hr />}
+        <div className="uic--mcg-subhead-2-text">
+          {isFetching ? (
+            <Spinner height="2rem" width="2rem" fill={colors.midnight} />
+          ) : (
+            children
```
RohitRajendran: Can you add text styles to make sure the text gets styled consistently?

RohitRajendran:
Percy won't be able to test the story properly since the description isn't visible by default

components/molecules/CardContentDescription/CardContentDescription.js
JamesIves: There's a class assigned on the div that enforces this - `uic--mcg-subhead-2-text`

components/molecules/CardContentDescription/CardContentDescription.js
JamesIves: There's now a `requiredIf` flag

JamesIves:
Added a `static text `story.

components/molecules/CardContentDescription/CardContentDescription.js
RohitRajendran: Should be left aligned https://app.zeplin.io/project/589b5c319e603a741d558f10/screen/5c7ee3dc760bbd51f354e11f

JamesIves: There's the `uic--mcg-subhead-2-text` class on the div that should enforce the actual styling.

Monique has deemed this text to be centered

components/molecules/CardContentDescription/CardContentDescription.spec.js
```diff
@@ -0,0 +1,96 @@
...
+    validate: () => true,
+    values: {
+      montezuma: 'montezuma',
+    },
+  };
+
+  const clock = useFakeTimers();
+  const component = mount(<CardContentDescription {...props} />);
+  component.instance().fetchData();
+  await clock.tick(1000);
```
CalMlynarczyk: You don't need to `await` this function; it doesn't return anything.

components/molecules/CardContentDescription/CardContentDescription.scss
```diff
@@ -0,0 +1,6 @@
+.uic--card-content-description {
+  .uic--card-content-description__content {
+    line-height: 1.4em;
```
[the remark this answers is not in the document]
JamesIves: Not sure how `em` managed to sneak its way in there.

components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -0,0 +1,122 @@
...
+
+  /** @inheritdoc */
+  componentDidUpdate(prevProps) {
+    if (this.props.validate && this.props.values) {
+      const prevValues = JSON.stringify(prevProps.values);
+      const values = JSON.stringify(this.props.values);
+      const valid = this.props.validate();
+
+      // Ensures the new values are different than the old ones.
+      if (values !== prevValues && valid) {
```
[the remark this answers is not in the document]
RohitRajendran: Is it really an issue? McGonagall currently take a similar approach to seeing if changes have been made in a card and I haven't seen any issues with it yet.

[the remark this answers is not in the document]
CalMlynarczyk: @RohitRajendran I would highly recommend we don't do it that way. It's an unreliable method for comparing objects, especially for non-serializable object types and nested objects.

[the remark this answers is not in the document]
CalMlynarczyk: @JamesIves That works. Also, we should put the existence check for `this.props.onChange` in the second-level `if` clause. Otherwise, if you don't include a handler the spinner shows indefinitely on a change.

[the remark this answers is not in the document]
RohitRajendran: We could say that values can only be one level deep and then do something like this so that you don't have to pass in a compare function.

```
const isDifferent = Object.keys(this.props.values).map((key) => {
  return this.props.values[key] === prevProps.values[key]
}).includes(false);
```

James, your idea of passing it in as an array instead of an object should also work right since the order should then be consistent?

I think avoiding having to passing in a compare function would be nice.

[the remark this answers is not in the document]
JamesIves: I like the array idea a little more because I feel like having three functions as props feels a bit convoluted. I'm open to the idea of having a compare function though.

PR #723, opened 2019-09-20
Card Content Description - Fetch on mount

RohitRajendran:
### Description
Fetches data on mount in card content description

### Testing Instructions
1. If valid, should check on mount


components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -33,26 +34,42 @@ class CardContentDescription extends Component {
...
-          valid,
-        });
-
-        this.fetchData();
-      } else if (!valid && valid !== this.state.valid) {
-        // If the field is invalid then the following state causes the text to hide.
-        this.setState({
-          valid,
-        });
+      if (values !== prevValues) {
```
CalMlynarczyk: The validation function won't fire now unless you change another one of the specified values. I think we have to always run the validation function like it was doing before. There is no guarantee that the validation function can't change without the `values` prop changing.

RohitRajendran: I still run the validation function but now I just check in the checkValidityAndFetch function

RohitRajendran: Ohhh I get what you mean

RohitRajendran: Fixed

PR #727, opened 2019-09-25
[APP-3203] Single-file FileInput for CSV Upload

labichn:
### Description
This PR adds a single-file FileInput component to the component library.

### Comments
- See [comps](https://zpl.io/2jWGe3p). Note: error border has been changed to 1px to match the standard Input.
- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-3203)


labichn:
I'm very open to feedback on component structure and CSS. The high-level styling approach follows the following article that @RohitRajendran sent me: https://benmarshall.me/styling-file-inputs.

components/atoms/FileInput/FileInput.js
```diff
@@ -0,0 +1,269 @@
+/** @module FileInput */
+import classNames from 'classnames';
+import PropTypes from 'prop-types';
+import React, {Component, createRef} from 'react';
+import {CardShellContext} from '../../molecules/CardShell/CardShell';
```
RohitRajendran:
```suggestion
import {CardShellContext} from '~components/molecules/CardShell/CardShell';
```

components/atoms/FileInput/FileInput.js
```diff
@@ -0,0 +1,269 @@
...
+              ref={this.inputNode}
+            >
+              <div
+                className="uic--mcgonagall-file-input__wrapper"
+                onDragLeave={this.onDragLeave}
+                onDragOver={this.onDragOver}
+                onDrop={this.onDrop}
+              >
+                <label className={labelClasses} htmlFor="uic--file-input-id">
+                  <div className={fakeButtonClasses}>
```
JamesIves: Are we not able to just use the `Button` component here instead? I don't think it needs to have either an `onChange` or a `to` prop to render the styles. If that's not applicable this should have the element name of `button` otherwise it's not obvious to screen readers that it's clickable.

components/atoms/FileInput/FileInput.js
```diff
@@ -0,0 +1,269 @@
...
+              style={style}
+              ref={this.inputNode}
+            >
+              <div
+                className="uic--mcgonagall-file-input__wrapper"
+                onDragLeave={this.onDragLeave}
+                onDragOver={this.onDragOver}
+                onDrop={this.onDrop}
+              >
+                <label className={labelClasses} htmlFor="uic--file-input-id">
```
JamesIves: You cannot tab or use keyboard shortcuts to trigger the file upload. I think changing the button element would solve this.

components/atoms/FileInput/FileInput.story.js
```diff
@@ -0,0 +1,96 @@
...
+stories
+  .addDecorator(withReadme(FileInputReadme))
+  .addDecorator(StateDecorator(store));
+
+store.subscribe(() => {
+  forceReRender();
+});
+
+/*
+export const inputMask = (defaultFunc) =>
```
JamesIves: No need to keep these comments here

components/atoms/FileInput/FileInput.story.js
```diff
@@ -0,0 +1,96 @@
...
+});
+
+const storyStyles = {
+  height: '100%',
+  maxWidth: '100%',
+  width: '80rem',
+  paddingTop: '3rem',
+};
+
+stories.add('default', () => (
```
JamesIves: Is there a way to mock the value of the file input via props? It would be good to have a story with a pre-filled file name in order for Percy to snapshot test it.

components/atoms/FileInput/FileInput.scss
```diff
@@ -0,0 +1,168 @@
...
+    background-color: transparent;
+  }
+
+  .uic--hidden-file-input {
+    border: 0;
+    clip: rect(0, 0, 0, 0);
+    height: 1px;
+    overflow: hidden;
+    padding: 0;
+    position: absolute !important;
```
JamesIves: Do we need `!important` here?  It may be better to switch the class using `classNames` if you need to override it.

components/atoms/FileInput/FileInput.scss
```diff
@@ -0,0 +1,168 @@
...
+  .uic--file-input-label {
+    @include source-sans-pro(2rem);
+    background-color: $white;
+    border: 1px solid $light-gray;
+    color: $midnight;
+    cursor: pointer;
+    display: inline-block;
+    margin-bottom: 0;
+    padding: 2.6rem 1.2rem 1rem 1rem;
+    width: 100%;
```
JamesIves: You can use `uic--w-100` here

components/atoms/FileInput/FileInput.scss
```diff
@@ -0,0 +1,168 @@
...
+    clip: rect(0, 0, 0, 0);
+    height: 1px;
+    overflow: hidden;
+    padding: 0;
+    position: absolute !important;
+    white-space: nowrap;
+    width: 1px;
+  }
+
+  .uic--file-input-label {
```
JamesIves: It would be nice to start structuring the CSS for this like some of the newer components, where we have `uic--file-input` and then `uic--file-input__label` - makes it easier to debug when searching for a class in the DOM. I'm not _too_ strongly in favor for this, but it would be nice to for consistency going forward. What do you think @RohitRajendran ?

components/atoms/FileInput/FileInput.scss
```diff
@@ -0,0 +1,168 @@
+$input-shadow: rgba(77, 0, 186, 0.5);
+$input-data-value-color: rgba(255, 255, 255, 0);
+
+.uic--mcgonagall-file-input {
+  line-height: normal;
```
JamesIves: Do these two properties do anything?

components/atoms/FileInput/FileInput.js
labichn: Fixed in 6028fb.

components/atoms/FileInput/FileInput.js
labichn: After discussion, @JamesIves and I agree that an actual HTML button is not desirable for accessibility reasons. A raw HTML file input does not have a button, despite appearing to have a button. This FileInput does not have a button, despite appearing to have a button.

components/atoms/FileInput/FileInput.js
labichn: You can, it's just not currently visible. Fixed in 70145c5.

components/atoms/FileInput/FileInput.story.js
labichn: Agreed. Fixed in 70145c5.

components/atoms/FileInput/FileInput.scss
labichn: No? Fixed in f333287.

components/atoms/FileInput/FileInput.scss
labichn: No? Fixed in cc12d6e.

components/atoms/FileInput/FileInput.scss
labichn: No? Fixed in cc12d6e.

components/atoms/FileInput/FileInput.scss
labichn: I'm glad to follow any kind of CSS naming convention--just let me know specifics.

labichn: Fixed in 2dc31ad.

components/atoms/FileInput/FileInput.story.js
labichn: Per discussion, we don't want to make a specific prop only for testing purposes.

PR #728, opened 2019-09-26
Display input appended character when no value is present 

CalMlynarczyk:
### Description
Display the appended character at the end of the input placeholder when one is provided. This includes appended characters added from masks. I also moved all of the masks into their own file so it's easier to work with the `Input` component.

### Testing Instructions
1. Go to the _percent_ story for `atom/Input`
2. "%" should display in the field while it is empty
3. Type a value into the field; the percent sign should still get appended to the value per the percentage mask


components/atoms/Input/masks.js
```diff
@@ -0,0 +1,208 @@
+import createAutoCorrectedDatePipe from 'text-mask-addons/dist/createAutoCorrectedDatePipe';
```
JamesIves: For consistency purposes I'd like to suggest naming this file `Input.masks.js`.

PR #729, opened 2019-09-30
[APP-4012] CardContentDescription Validation

JamesIves:
Adds some additional validation to the `CardContentDescription` component so it now invalidates the McG card when the fetch status is in flight. Also added a prop called `isValid` which will essentially hide the component if a criteria isn't met. This last part is needed in some circumstances where we always need to make a fetch request but not show the component.  (See auto loan flow)

* Enter input in the box in Storybook. 
* If the component is fetching the continue button will be greyed out.

I don't love this solution, I'm open to suggestions if you can think of something better.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4012)


components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -99,15 +107,15 @@ class CardContentDescription extends Component {
 
     return (
       <div className={containerClasses} style={style}>
-        {!removeTopBorder && <hr />}
+        {((!removeTopBorder && isValid) || isFetching) && <hr />}
```
CalMlynarczyk: I think this should be `{(!removeTopBorder && (isValid || isFetching))...` so that we're not ignoring the `removeTopBorder` prop while fetching.

components/molecules/CardContentDescription/CardContentDescription.js
```diff
@@ -99,15 +107,15 @@ class CardContentDescription extends Component {
...
         <div className="uic--card-content-description__content uic--text-center uic--mcg-subhead-2-text">
           {isFetching ? (
             <Spinner height="2rem" width="2rem" fill={colors.midnight} />
           ) : (
-            children
+            isValid && children
           )}
         </div>
-        {!removeBottomBorder && <hr />}
+        {((!removeBottomBorder && isValid) || isFetching) && <hr />}
```
CalMlynarczyk: Same thing about ignoring that prop here.

JamesIves:
@CalMlynarczyk We don't - I'm going to remove `validate` in favor for `isValid` which will just hide the component when the criteria isn't correct for it to be shown. Preventing it from firing `onChange` doesn't make sense when we're primarily using the data returned from it for form validation.

PR #748, opened 2019-10-07
Input z-index shift

JamesIves:
Really small z-index shift on the input tooltip as it's overlapping the modal right now. Nothing really should be higher than 1500. 

* Look at the `input` component. 
* Highlight it, and trigger a modal in McG.
* The input outline should not be visible over the modal. 


labichn:
z-index
Want to take care of DropDown, Cabinet, and Modal at the same time?

PR #750, opened 2019-10-08
Storybook DocsPage/MDX Support

JamesIves:
This branch introduces a proof of concept for Storybook's new DocsPage update which you can view [here](https://medium.com/storybookjs/storybook-docspage-e185bc3622bf).

Screen Shot 2019-10-08 at 10 07 10 AM

There's a few big changes in here.

* `Component.story.js` and `Component.md` are now merged into one, and now have a single unified `Component.story.mdx` file. These files define the documentation and the stories in one file.
* `addon-info` has  been removed as it's now obsolete. `addon-readme` will soon follow once all stories have been converted.
* Stories are defined inline using a set of provided components by Storybook. You can view a preview of what this looks like here with the `<Box />` component.
* Prop tables are generated using `<Props for={Component} />` inside the `mdx` file.
* I've updated the linting rules, along with the generator to support `mdx`.\
* Added some additional UI colors to the theme.

In this pull request I've updated the `Box`, `BubbleSelector` and the `BubbleDropdown` components to support this new format. To view the changes you can click on the story in Storybook and click on the `Docs` tab at the top to view the new examples.

#### Known Issues

There's a couple of known issues with this.

* Using addon-knobs with a helper function to generate props results in the actual function call being rendered in the code preview instead of the computed values. I've placed an issue with the Storybook team on this here: https://github.com/storybookjs/storybook/issues/8342
* For whatever `forceReRender` isn't needed anymore for stateful components. Not sure if there will be any other side effects from that later down the line or not.

* Load up a story that is supporting this new format and click the `Docs` tab. This tab will show for any story.
* All old stories using the old format of `Component.story.js` will continue to function, but should no longer be used.

I'd like to update all of the old stories into this format, and I predict it will take around an afternoon to complete this work if everyone is on board. I think this way of writing documentation is much cleaner.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4006)


JamesIves:
Some additional reading as to why I'm going to hold off merging this for now. Once we get a fix for this in place I feel confident in migrating to the new format of mdx stories: https://github.com/storybookjs/storybook/issues/8342#issuecomment-540696692

components/atoms/Box/Box.story.mdx
```diff
@@ -0,0 +1,62 @@
...
+  description="Montezuma is honestly the best cat I've ever seen."
+  icon={CatIllustration}
+/>
+```
+
+<Preview>
+  <Story name="default">
+    <Box {...defaultProps(false)} />
+  </Story>
+</Preview>
```
RohitRajendran: Since the Preview has a show code button, I don't think it's necessary to have a separate code example as long as the preview code is done properly.

components/atoms/Box/Box.story.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+### Using an Icon
+
+You can include an icon or illustration at the top of the box by passing it in using the `icon` prop. When the `selected` prop is set to `true` the box will also pass the value to the `illuminate` prop on the icon component, which will trigger the activated state for most of the illustrations found in this library.
+
+<Preview>
+  <Story name="selected">
+    <Box {...defaultProps(true)} icon={CarCircleIllustration} />
```
RohitRajendran: I think we should stop using the default props so that the code examples display properly

components/atoms/BubbleSelector/BubbleSelector.story.mdx
```diff
@@ -0,0 +1,75 @@
...
+/>
+```
+
+<Preview>
+  <Story name="dynamic width">
+    <BubbleSelector {...defaultProps({})} />
+  </Story>
+</Preview>
+
+You can apply a fixed width to the component using the `fixedWidth` prop.
```
RohitRajendran: Not really specific to this PR but I think we should add when/why you would use the fixedWidth prop

components/atoms/BubbleSelector/BubbleSelector.story.mdx
```diff
@@ -0,0 +1,75 @@
...
+
+<Preview>
+  <Story name="fixed width">
+    <BubbleSelector {...defaultProps({fixedWidth: '17rem'})} />
```
RohitRajendran: This isn't looking any different than the dynamic width

components/atoms/Box/Box.story.mdx
RohitRajendran: Since we can't not use defaultProps, I think we should show the code example for every preview

PR #768, opened 2019-10-10
[APP-3973] Create data table component

CalMlynarczyk:
### Description
Create a data table component capable of displaying both tabular data in a traditional table layout, and an expandable layout with variable sub-components.

### Testing Instructions
Go to /?path=/story/complex-organisms-datatable--default

### Comments
This was originally going to consist of a lot more, including the expandable section for the new cashflow table. I ultimately decided that should be part of client and the caller can inject it into the table using the `SubComponent` prop and a custom renderer. That table style appears to only be a one-off, so trying to generalize it is pointless for now until we find a reason to do so.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-3973)


components/atoms/icons/StarCircleIcon/StarCircleIcon.story.js
```diff
@@ -0,0 +1,19 @@
...
+import StarCircleIcon from './StarCircleIcon';
+import StarCricleIconReadme from './StarCircleIcon.md';
+
+const stories = storiesOf('Atoms/Icons/StarCircleIcon', module);
+
+stories.addDecorator(withReadme(StarCricleIconReadme));
+
+const defaultProps = () => ({
+  fill: text('fill', '#FFF'),
+  outerFill: text('outer fill', 'Green'),
```
JamesIves:
```suggestion
  outerFill: text('outerFill', 'Green'),
```

Can we keep the prop name in the knobs table consistent with what's being passed into the component?

package.json
```diff
@@ -86,6 +86,7 @@
     "react-required-if": "*",
     "react-select": "*",
     "react-spring": "*",
+    "react-table": "6",
```
JamesIves: Should this be `=6`?

CalMlynarczyk: Both formats produce the same result.

PR #769, opened 2019-10-10
Release October 10th

CalMlynarczyk:
* [ ] (In Progress) [APP-3973 - Replace the cash flow cards with the cash flow table](https://unitedincome.atlassian.net/browse/APP-3973)


JamesIves:
@CalMlynarczyk I think that version of rc-slider is going to cause some issues. I reverted it yesterday after updating but forgot to lock the version number. Percy may flag it

PR #784, opened 2019-10-17
Add custom edit card link text to simple summary

labichn:
### Description
This PR allows customization of the "Edit" link/button text in the SimpleSummary component.

### DIFF
--- a/components/molecules/CardSummaries/SimpleSummary/SimpleSummary.js
+++ b/components/molecules/CardSummaries/SimpleSummary/SimpleSummary.js
@@ -6,7 +6,7 @@ import Button from '~components/atoms/Button/Button';
 import {exclusive} from '~proptypes';
 import '../CardSummaries.scss';
 
-const SimpleSummary = ({shortTitle, editCard, answers, to}) => {
+const SimpleSummary = ({shortTitle, editCard, editCardText, answers, to}) => {
   return (
     <div className="uic--card-summary uic--container-fluid">
       <div className="uic--row">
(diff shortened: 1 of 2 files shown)


components/molecules/CardSummaries/SimpleSummary/SimpleSummary.js
```diff
@@ -16,7 +16,7 @@ const SimpleSummary = ({shortTitle, editCard, answers, to}) => {
         <div className="uic--col-6 uic--col-sm-1 uic--order-sm-last">
           <div className="uic--d-flex uic--justify-content-end">
             <Button variant="link" onClick={editCard} to={to}>
-              Edit
+              {editCardText || 'Edit'}
```
RohitRajendran: Let's just use prop defaults to default editCardText to default

labichn: Agreed, and fixed in 16ed9f5.

components/molecules/CardSummaries/SimpleSummary/SimpleSummary.js
```diff
@@ -39,10 +39,14 @@ SimpleSummary.propTypes = {
   answers: PropTypes.arrayOf(PropTypes.string).isRequired,
   /** Handler called to edit the card, used instead of `to` for McGonagall. */
   editCard: and([PropTypes.func, exclusive(['to'])]),
+  /** Text to display in the edit button--defaults to 'Edit' if not present */
+  editCardText: PropTypes.string.isRequired,
```
RohitRajendran: This shouldn't be required

RohitRajendran: Nevermind, didn't realize that default prop makes it so that it doesn't error when a value isn't passed in.

labichn: @RohitRajendran and I discussed, and I ran it by @JamesIves as well. The error message is more informative if the default is removed in some future commit (refers to `editCardText` directly), otherwise the error message yells about undefined children passed to `ButtonInterior`. For this reason, it makes sense to leave the `editCardText` prop as required.

PR #792, opened 2019-10-24
[APP-4006] Storybook MDX Docs

JamesIves:
Migrates _all_ component library docs to `mdx` format. I've cleaned up a lot of the documentation, and added a prop table to each one. There's likely a few spelling errors, but once this merges we can fix them as we see them. The generator has also been updated to support this new format, along with the linter/formatter.

I've disabled `addon-info` and `addon-readme` as part of this as they are no longer required. I've also ran the `Organize Import` tool that VSCode has on all `.js` files in the repo as a general cleanup item. At some point in the future I'll figure out a way to lint that, but for now we can manually do it.

Stories and docs are still defined separately for now, and code samples are provided beneath the story. Eventually we may change how this works based on how DocPages evolves but this is still a step in the right direction for now. 

* Load storybook.
* Click on the `Docs` tab.
* Click through each story and observe the new docs.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4006)


components/atoms/Box/Box.mdx
```diff
@@ -0,0 +1,59 @@
+import {Preview, Props} from '@storybook/addon-docs/blocks';
+import Box from './Box';
+import CarCircleIllustration from '~components/atoms/illustrations/CarCircleIllustration/CarCircleIllustration';
+
+# Box
+
+Displays content within a simple box with an icon, title, and description. The `Box` component is primary used in the McGonnagal flow selection screens.
```
RohitRajendran:
```suggestion
Displays content within a simple box with an icon, title, and description. The `Box` component is primarily used in the McGonnagal flow selection screens.
```

components/atoms/Box/Box.mdx
```diff
@@ -0,0 +1,59 @@
...
+import Box from '@unitedincome/components/dist/Box';
+```
+
+```javascript
+import {Box} from '@unitedincome/components';
+```
+
+## Example 🚀
+
+<Preview>
```
RohitRajendran: Missing code examples

components/atoms/Button/Button.mdx
```diff
@@ -16,13 +19,23 @@ import {Button} from '@unitedincome/components';
...
-<Button variant="primary" light={false} dark={true} disabled={false}>
+<Button variant="primary" light={true} dark={false} disabled={false}>
   Submit
 </Button>
 ```
 
-## Using with React Router
+### Variants
+
+There's multiple different button variants which can be utilized, along with `light` and `dark` props to adjust how the buttons are displayed. For a full list of combinations please refer to the prop table and the provided stories.
```
RohitRajendran: I feel like it would be good to have the variants here so that someone can easily copy and paste code

JamesIves: That's fair, I can add a couple

JamesIves: Done!

components/atoms/Button/Button.mdx
```diff
@@ -0,0 +1,98 @@
...
+
+# Button
+
+The Button component displays a HTML button.
+
+## Importing 📦
+
+You can use this component using one of the following import patterns.
+
+```javascript
```
RohitRajendran: Since client uses a wrapper, should we include a note about importing it from there if you're using doing this in client?

JamesIves: I don't think so as this is open source. I like taking the neutral approach with this library, and having additional internal docs for this sort of thing.

PR #811, opened 2019-10-25
Set explicit versions for dependencies in package.json

CalMlynarczyk:
### Description
Recent versions of Yarn no longer handle the `*` version of packages in the CLI. Also, we had some dislike on the team as to using these because it made regenerating the lockfile a pain. Use exact numeric versions in `package.json` for dependencies.


package.json
```diff
@@ -27,92 +27,92 @@
...
+    "accounting": "0.4.1",
+    "airbnb-prop-types": "2.15.0",
+    "babel-loader": "8.0.6",
+    "babel-plugin-istanbul": "5.2.0",
+    "babel-plugin-module-resolver": "3.2.0",
+    "bootstrap": "4.3.1",
+    "classnames": "2.2.6",
+    "core-js": "3.3.4",
+    "css-loader": "3.2.0",
+    "d3-array": "1.2.4",
```
RohitRajendran: Can we keep these with the = so that we know we wanted to version lock it?

```suggestion
    "d3-array": "=1.2.4",
```

CalMlynarczyk: We can try and see if Dependabot version-locks packages with this format, but I don't know if it will work. Yarn and NPM treat the [`=` operator on a version as optional](https://yarnpkg.com/en/docs/dependency-versions#toc-comparators), so these two formats are effectively the same.

Dependabot has its own internal mechanisms for version-locking using commands against PRs for those packages. Moving away from the "lockfile-only" update approach would require us to use those mechanisms instead since it isn't really going to care what version format we use in `package.json` anymore.

CalMlynarczyk: Fair enough. I've indicated the locked versions with the suggested syntax.

PR #825, opened 2019-10-29
Changes publishing pipeline

JamesIves:
Changes the publishing pipeline so `develop` goes into `/develop`, and `master` sits int he root.

I guess we just run it and find out


publishStorybook.sh
```diff
@@ -24,6 +24,14 @@ echo "Bucket: ${BUCKET}"
...
+# Publish 'master' to the root directory', and other deployment branches into their own folder.
+if [[ $BRANCH == *"master"* ]]; 
+then
+    aws s3 sync $STORYBOOK_DIR s3://$DOCS_BUCKET --delete
+else
+    aws s3 sync $STORYBOOK_DIR s3://$DOCS_BUCKET/$BRANCH --delete
+fi
+
+aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --path "/*"
 aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --path "/${BRANCH}/*"
```
JamesIves: I'm wondering if this will invalidate everything in the /develop/* folder, resorting in the other one being unnecessary. I've never used this cli before.

publishStorybook.sh
```diff
@@ -24,6 +24,14 @@ echo "Bucket: ${BUCKET}"
 
 PROFILE=${AWS_PROFILE:-$PROFILENAME}
 STORYBOOK_DIR=.build/
-aws s3 sync $STORYBOOK_DIR s3://$DOCS_BUCKET/$BRANCH --delete
 
+# Publish 'master' to the root directory', and other deployment branches into their own folder.
+if [[ $BRANCH == *"master"* ]]; 
+then
+    aws s3 sync $STORYBOOK_DIR s3://$DOCS_BUCKET --delete
```
matthewheck: So, this will delete everything in the bucket, wiping out the develop branch.

publishStorybook.sh
matthewheck: This is just invalidation of the cloudfront cache. All it does is tell cloudfront to act as if everything in the target folder is brand new and reset the cache.

So, you won't need ln 37 (invaliding by branch), as ln 36 will handle it.

publishStorybook.sh
JamesIves: Is it the `--delete` flag that does this?

JamesIves: Ok I may have a solution that will work. There's a `--exclude` option apparently.

publishStorybook.sh
```diff
@@ -24,6 +24,13 @@ echo "Bucket: ${BUCKET}"
 
 PROFILE=${AWS_PROFILE:-$PROFILENAME}
 STORYBOOK_DIR=.build/
-aws s3 sync $STORYBOOK_DIR s3://$DOCS_BUCKET/$BRANCH --delete
 
-aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION" --path "/${BRANCH}/*"
+# Publish 'master' to the root directory', and other deployment branches into their own folder.
+if [[ $BRANCH == *"master"* ]];
```
matthewheck: wondering if we want to make this specifically "master" without the wildcards. For instance, this might still get triggered by a branch called "remastered".

publishStorybook.sh
JamesIves: sure, we can change it if it doesn't work

PR #828, opened 2019-10-29
PrinterIcon and PrinterCircleIcon

JamesIves:
Adds the `PrinterIcon` and the `PrinterCircleIcon`.

* It's an icon, just look at it!


components/atoms/icons/TwitterIcon/TwitterIcon.js
```diff
@@ -1,12 +1,12 @@
-/** @module TwitterIcon */
+/** @module PrinterIcon */
```
RohitRajendran: Why'd you change things in the TwitterIcon?

JamesIves: woops

JamesIves: fixed

PR #867, opened 2019-11-19
[APP-4064] Account Box Cards Draft

angelojcruzjr:
Created two new components, one is an `atom` called `CardContainer` and the other is a `molecule` titled `CardContainerWithRows`.

Run storybook and checkout the components! Also, more importantly, peep the code for how I set these up 👀.

Very open to changing the names of these, just put in whatever popped into my head  😅

**Note:**
- No tests have been written yet
- No component interactivity yet

Want to make sure these are on the right track before I continue! Original comps can be found in the Jira ticket.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/APP-4064)


components/atoms/CardContainer/CardContainer.js
```diff
@@ -0,0 +1,57 @@
...
+      cardContent: props.cardContent,
+      isUiAccount: props.isUiAccount,
+    };
+  }
+
+  /** @inheritdoc */
+  render() {
+    const {title, cardContent, isUiAccount} = this.props;
+
+    let accountClass = 'uic--card-container-title ';
```
RohitRajendran: Check out the classnames package that we use in other components like the FormSummary. It makes it much cleaner to do conditional classes

components/atoms/CardContainer/CardContainer.js
```diff
@@ -0,0 +1,57 @@
...
+      </div>
+    );
+  }
+}
+
+CardContainer.propTypes = {
+  /** The message to be displayed */
+  title: PropTypes.string.isRequired,
+  /** Is this a UI account? */
+  isUiAccount: PropTypes.bool,
```
RohitRajendran: Instead of a toggle to change the image, I think we need to let this be flexible for whatever image that wants to be used.

components/atoms/CardContainer/CardContainer.js
```diff
@@ -0,0 +1,57 @@
+/** @module Ranking */
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
+import './CardContainer.scss';
+
+/** Renders a list which allows  */
+class CardContainer extends Component {
```
RohitRajendran: I think we should rename this so that it doesn't get confused with the cards for the McG flow. I'm not really sure what though, maybe InfoBox?

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
+/** @module Ranking */
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
```
RohitRajendran:
```suggestion
import React, {Component, Fragment} from 'react';
```

So that you can just do `Fragment` instead of `React.Fragment`.

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
...
+      </React.Fragment>
+    );
+  }
+}
+
+CardContainerWithRows.propTypes = {
+  rowData: PropTypes.arrayOf(
+    PropTypes.shape({
+      /** Key for this piece of data */
+      key: PropTypes.string.isRequired,
```
RohitRajendran: We should make this optional and just use label as the key if none provided

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
...
+      /** Label for this piece of data */
+      label: PropTypes.string.isRequired,
+      /** Value for this piece of data */
+      value: PropTypes.number.isRequired,
```
RohitRajendran: Value should be a string. We should get rid of the format object because the value should be the preformatted string

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
...
+  ).isRequired,
+  /** Optional formatting */
+  format: PropTypes.shape({
+    /** Prepend this to our data */
+    prepend: PropTypes.string,
+    /** Append this to our data */
+    append: PropTypes.string,
+  }),
+  /** Optional override statement */
+  override: PropTypes.shape({
```
RohitRajendran: Lets call this the footer and allow JSX to be passed through

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
...
+          </div>
+        ))}
+        {override ? this.generateOverride() : ''}
+      </React.Fragment>
+    );
+  }
+}
+
+CardContainerWithRows.propTypes = {
+  rowData: PropTypes.arrayOf(
```
JamesIves: Missing prop comment here

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
...
+    PropTypes.shape({
+      /** Key for this piece of data */
+      key: PropTypes.string.isRequired,
+      /** Label for this piece of data */
+      label: PropTypes.string.isRequired,
+      /** Value for this piece of data */
+      value: PropTypes.number.isRequired,
+    }).isRequired,
+  ).isRequired,
+  /** Optional formatting */
```
JamesIves: Can you make these comments a bit more descriptive across the board? Optional formatting for what?

components/molecules/CardContainerWithRows/CardContainerWithRows.scss
```diff
@@ -0,0 +1,28 @@
+$container-min-height: 2.4rem;
+$override-line-top: 1.2rem;
+
+.uic--card-container-with-rows {
+  align-items: center;
+  display: flex;
```
JamesIves: There's a series of helper classes you can use for these, ie `uic--d-flex` etc.

components/atoms/CardContainer/CardContainer.story.js
```diff
@@ -0,0 +1,34 @@
...
+  overflow: 'none',
+  textOverflow: 'ellipsis',
+};
+
+stories
+  .addDecorator(withKnobs)
+  .add('default', () => (
+    <CardContainer
+      title={text('Text', 'My 401k Account')}
+      isUiAccount={boolean('IsUiAccount', false)}
```
JamesIves: Can you use the `defaultProps` function model for this?

components/atoms/CardContainer/CardContainer.scss
```diff
@@ -0,0 +1,47 @@
...
+  .uic--card-container-title {
+    background-color: $midnight;
+    background-position: 100% 0%;
+    background-repeat: no-repeat;
+    background-size: $background-size;
+    min-height: $min-height;
+    position: relative;
+
+    // so that the child node does not cause
+    // the parents node onMouseOut event to fire
```
JamesIves: Does this need to be a wildcard though? I'd prefer if we wrapped it in a class and assigned the property there. This isn't a pattern I'm fond of.

components/atoms/CardContainer/CardContainer.js
JamesIves: I agree with @RohitRajendran  - I'm not a fan of naming this `isUiAccount` either, as this library is open sourced we should be using agnostic language here, and throughout the whole component.

components/atoms/CardContainer/CardContainer.js
```diff
@@ -0,0 +1,57 @@
...
+  /** The message to be displayed */
+  title: PropTypes.string.isRequired,
+  /** Is this a UI account? */
+  isUiAccount: PropTypes.bool,
+  /** The (optional) cardContent JSX to be displayed */
+  cardContent: PropTypes.node.isRequired,
+};
+
+CardContainer.defaultProps = {
+  isUiAccount: true,
```
JamesIves: Same comment about the language. Should this be defaulted to true?

components/atoms/CardContainer/CardContainer.js
```diff
@@ -0,0 +1,57 @@
...
+import './CardContainer.scss';
+
+/** Renders a list which allows  */
+class CardContainer extends Component {
+  /** @inheritdoc */
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      title: props.title,
```
JamesIves: Why do we need to bind `props.title` to state here?  It looks like you can just rely on props and make this a functional component.

components/atoms/CardContainer/CardContainer.scss
```diff
@@ -0,0 +1,47 @@
...
+      right: 6rem;
+      text-align: left;
+      top: 2.5rem;
+    }
+
+    &.non-ui-account {
+      background-image: url('../../../assets/icons/investment.svg');
+    }
+
+    &.ui-account {
```
JamesIves: Same comment regarding language.

components/atoms/CardContainer/CardContainer.scss
```diff
@@ -0,0 +1,47 @@
...
+      color: $white;
+      left: 2rem;
+      position: absolute;
+      right: 6rem;
+      text-align: left;
+      top: 2.5rem;
+    }
+
+    &.non-ui-account {
+      background-image: url('../../../assets/icons/investment.svg');
```
JamesIves: These svgs should probably be passed into the component as opposed to hardcoding them. What if the user is using this in an application other than ui?

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
+/** @module Ranking */
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
+import './CardContainerWithRows.scss';
+
+/**
+ * Utility function that formats our values
```
JamesIves: Confused about what this does by this comment. Which values?

components/molecules/CardContainerWithRows/CardContainerWithRows.js
```diff
@@ -0,0 +1,120 @@
...
+  returnVal += val;
+
+  if (formatObj.append) {
+    returnVal += formatObj.append;
+  }
+
+  return returnVal;
+}
+
+/** Renders a list which allows  */
```
JamesIves: Incomplete comment

components/molecules/CardContainerWithRows/CardContainerWithRows.scss
RohitRajendran: I still don't agree with using helper classes if you anyways need to use a custom class

JamesIves: 🥊

components/atoms/CardContainer/CardContainer.js
angelojcruzjr: No longer an issue since we don't ask for isUiAccount anymore 👀

components/atoms/CardContainer/CardContainer.js
angelojcruzjr: Updated logic in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960 so that we just accept an image URL

components/atoms/CardContainer/CardContainer.js
angelojcruzjr: Renamed to InfoBox in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Updated to use just Fragment in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Updated to be optional in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Updated to use string values that the user passes in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Renamed to footer in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Added in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Went in and re-did all the comments in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.scss
angelojcruzjr: I used a helper class in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960 ...sorry @RohitRajendran 🙈

components/atoms/CardContainer/CardContainer.story.js
angelojcruzjr: Updated both components to use `defaultProps` function model in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/atoms/CardContainer/CardContainer.scss
angelojcruzjr: I just left this in the child class in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960. I think it makes more sense since we control what goes inside here so we don't necessarily need to account for all descending children.

components/atoms/CardContainer/CardContainer.js
angelojcruzjr: No longer applicable due to logic changes in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/atoms/CardContainer/CardContainer.js
angelojcruzjr: You're right I wasn't using state for anything in either component. Updated both to be pure components in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/atoms/CardContainer/CardContainer.scss
angelojcruzjr: No longer applicable due to logic changes in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/atoms/CardContainer/CardContainer.scss
angelojcruzjr: No longer applicable due to logic changes in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Updated all comments in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

components/molecules/CardContainerWithRows/CardContainerWithRows.js
angelojcruzjr: Updated all comments in https://github.com/UnitedIncome/components/pull/867/commits/21c1882106753193b3b995f2e26d6b5578cca960

angelojcruzjr:
@RohitRajendran @JamesIves made a bunch of updates if you want to take another pass at this whenever y'all have free time. Going to work on tests in the meantime!

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,34 @@
...
+  .uic--info-box-title-area {
+    background-color: $midnight;
+    background-position: 100% 0%;
+    background-repeat: no-repeat;
+    background-size: $background-size;
+    min-height: $min-height;
+    position: relative;
+
+    .uic--info-box-title-text {
+      @extend %uic--header-2-text;
```
RohitRajendran: This should just get applied as a class instead of extending it.

components/atoms/InfoBox/InfoBox.story.js
```diff
@@ -0,0 +1,53 @@
...
+store.subscribe(() => {
+  forceReRender();
+});
+
+const defaultProps = ({formName, title, imageURL, infoBoxContent}) => ({
+  name: formName,
+  title: text('title', title),
+  imageURL: text('imageURL', imageURL),
+  infoBoxContent: object('infoBoxContent', infoBoxContent),
+  onChange: (name, value) => store.set({[name]: value}),
```
RohitRajendran: What do you need the onChange for?

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
```diff
@@ -0,0 +1,48 @@
+/** @module InfoBoxFormattedRows */
+import PropTypes from 'prop-types';
+import React, {Component, Fragment} from 'react';
+import './InfoBoxFormattedRows.scss';
+
+/** Renders an InfoBox with the title formatted the same way the
+ *  basic title us, but with a well-defined row structure where data
```
RohitRajendran: `basic title us` typo

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
```diff
@@ -0,0 +1,48 @@
...
+        {rowData.map((row) => (
+          <div
+            className="uic--info-box-formatted-row uic--d-flex uic--justify-content-between"
+            key={row.key ? row.key : row.label}
+          >
+            <div className="uic--body-text">{row.label}</div>
+            <div className="uic--mcg-subhead-text">{row.value}</div>
+          </div>
+        ))}
+        {footer ? footer : ''}
```
RohitRajendran: If the user is passing in a footer, the divider line thing should get added. That shouldn't be something the user has to include

components/molecules/CardContainerWithRows/CardContainerWithRows.scss
RohitRajendran: How could you @angelojcruzjr!? 💔

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
RohitRajendran: We might want to actually make this functionality for the basic InfoBox

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,38 @@
...
+      </div>
+    );
+  }
+}
+
+InfoBox.propTypes = {
+  /** The main title that will displayed near the top of our InfoBox */
+  title: PropTypes.string.isRequired,
+  /** The infoBoxContent JSX to be displayed below the title */
+  infoBoxContent: PropTypes.node.isRequired,
```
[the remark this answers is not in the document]
angelojcruzjr: Changed to `content` in https://github.com/UnitedIncome/components/pull/867/commits/f44defbd99fd470248924192d3a00b8791ce5ad6

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,34 @@
+$min-height: 10rem;
+$background-size: 17rem;
+
+.uic--info-box {
+  @include antialias;
```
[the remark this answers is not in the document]
angelojcruzjr: Ah I see - this makes sense. It was not rendering to spec based on the comps I was seeing but that's because we will antialias on the app itself and that should be good to go. Changed in https://github.com/UnitedIncome/components/pull/867/commits/f44defbd99fd470248924192d3a00b8791ce5ad6

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: Done and done in https://github.com/UnitedIncome/components/pull/867/commits/f44defbd99fd470248924192d3a00b8791ce5ad6

components/atoms/InfoBox/InfoBox.story.js
angelojcruzjr: I need to read over these storybook docs in detail 😞 - thought it was needed for component interactivity but turns out that's not the case. Got rid of it in https://github.com/UnitedIncome/components/pull/867/commits/f44defbd99fd470248924192d3a00b8791ce5ad6

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
angelojcruzjr: More better grammar in https://github.com/UnitedIncome/components/pull/867/commits/f44defbd99fd470248924192d3a00b8791ce5ad6

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
angelojcruzjr: Added in the line and made the footer a part of the broader `InfoBox` component in https://github.com/UnitedIncome/components/pull/867/commits/f44defbd99fd470248924192d3a00b8791ce5ad6

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,38 @@
...
+        <div className="uic--info-box-title-area" style={cardStyle}>
+          <div className="uic--info-box-title-text">{title}</div>
+        </div>
+        <div className="uic--info-box-information-area">{infoBoxContent}</div>
+      </div>
+    );
+  }
+}
+
+InfoBox.propTypes = {
```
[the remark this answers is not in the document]
angelojcruzjr: Working on this still! Watch for the next commit.

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.mdx
```diff
@@ -0,0 +1,82 @@
...
+import InfoBoxFormattedRows from '@unitedincome/components/dist/InfoBoxFormattedRows';
+```
+
+```javascript
+import {InfoBoxFormattedRows} from '@unitedincome/components';
+```
+
+## Example 🚀
+
+_Note:_ The imageURL should link to whatever custom image you would like to pass into the InfoBox.
```
JamesIves: This is super nitpicky, but can we just add this as a prop type comment? There's no real need to add a note here when it will show up in the table below.

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
```diff
@@ -0,0 +1,45 @@
+/** @module InfoBoxFormattedRows */
+import PropTypes from 'prop-types';
+import React, {Component, Fragment} from 'react';
+
+/** Renders an InfoBox with the title formatted the same way the
+ *  basic title is, but with the content being a well-defined row
+ *  structure where data labels are on the left and their associated
+ *  values on the right.
+ */
+class InfoBoxFormattedRows extends Component {
```
JamesIves: This could  be a functional component as you're using none of the reacts class component lifecycle methods besides render, ie `const InfoBoxFormattedRows = () => ( <div></div>)`

components/atoms/InfoBox/InfoBox.mdx
```diff
@@ -0,0 +1,45 @@
...
+import InfoBox from '@unitedincome/components/dist/InfoBox';
+```
+
+```javascript
+import {InfoBox} from '@unitedincome/components';
+```
+
+## Example 🚀
+
+_Note:_ The imageURL should link to whatever custom image you would like to pass into the InfoBox.
```
JamesIves: Same note about the readme

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,55 @@
+/** @module Ranking */
+import PropTypes from 'prop-types';
+import React, {Component, Fragment} from 'react';
+import './InfoBox.scss';
+
+/** Renders a basic card with a title component on top and
+ *  description/more information below the title portion.
+ */
+class InfoBox extends Component {
```
JamesIves: Same comment about the other functional component

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,55 @@
...
+      <div className="uic--info-box">
+        <div
+          className="uic--info-box-title-area"
+          style={imageURL ? cardStyle : ''}
+        >
+          <div className="uic--info-box-title-text uic--header-2-text">
+            {title}
+          </div>
+        </div>
+        <div className="uic--info-box-information-area">
```
JamesIves: If this is the footer of the component, using the `<footer>` element may be more appropriate.

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,55 @@
+/** @module Ranking */
```
JamesIves:
```suggestion
/** @module InfoBox */
```

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,44 @@
+$min-height: 10rem;
+$background-size: 17rem;
+$container-min-height: 2.4rem;
+$border-top-offset: ($container-min-height / 2);
+
+.uic--info-box {
+  border: 1px solid $light-gray;
+  cursor: default;
```
JamesIves: What is the need for `cursor: default` here?

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,55 @@
...
+  /** @inheritdoc */
+  render() {
+    const {title, content, imageURL, footer} = this.props;
+
+    const cardStyle = {
+      backgroundImage: `url(${imageURL})`,
+    };
+
+    return (
+      <div className="uic--info-box">
```
JamesIves: `<aside>` maybe? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,55 @@
...
+        <div
+          className="uic--info-box-title-area"
+          style={imageURL ? cardStyle : ''}
+        >
+          <div className="uic--info-box-title-text uic--header-2-text">
```
JamesIves: I feel like this should be some kind of header tag instead of a div as it defines content.

assets/icons/ui-account.svg
```diff
@@ -0,0 +1,13 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="265" height="160" viewBox="0 0 265 160">
```
JamesIves: If this isn't actually used for anything let's remove it and place a hardcoded link to an image for the previews.

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,44 @@
...
+    background-size: $background-size;
+    min-height: $min-height;
+    position: relative;
+
+    .uic--info-box-title-text {
+      color: $white;
+      left: 2rem;
+      pointer-events: none;
+      position: absolute;
+      right: 6rem;
```
JamesIves: I can't seem to figure out what this does

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,44 @@
...
+      text-align: left;
```
JamesIves: Probably don't need to define the text alignment here

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,55 @@
...
+            {title}
+          </div>
+        </div>
+        <div className="uic--info-box-information-area">
+          {content}
+          {footer ? (
+            <Fragment>
```
RohitRajendran: You can just do `footer && (` instead of the ternary

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,55 @@
...
+              <div className="uic--info-box-formatted-row border-top" />
```
RohitRajendran: Why not an hr?

angelojcruzjr: Updated in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: 🤦🏽‍♂️true. Updated in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: Correct - extraneous here. Got rid of it in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

assets/icons/ui-account.svg
angelojcruzjr: Removed and replaced with a hardcoded svg link in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Changed to header in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: So I looked into the docs and from what I could tell asides are supposed to be used for tangentially related information to the main page...are you saying that this entire InfoBox concept would be an aside? I can see an argument for this. Just want to confirm before I change. Would like to hear @RohitRajendran 's opinion as well 😇

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Woops. Updated in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: Leftover from a copy and paste. Got rid of it in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Yup better semantics. Changed in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Made functional in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.mdx
angelojcruzjr: Updated readme in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
angelojcruzjr: Updated to be functional in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.mdx
angelojcruzjr: Updated README in https://github.com/UnitedIncome/components/pull/867/commits/0e081efa568fc65b91be2baff1a8ba08819e8f5f

components/atoms/InfoBox/InfoBox.js
JamesIves: Yeah, as it contains content which is related to the page but not the sole purpose of the page I feel it may make sense in this scenario. `<section />` may also be applicable for these. 

I'm not attached to either idea, just a possible area of improvement.

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: So this positions the `title` to the position it is currently in

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Made it `<aside />` in https://github.com/UnitedIncome/components/pull/867/commits/13a29992ae9f227b4d94ca78a84cefb49ff437b0

components/atoms/InfoBox/InfoBox.js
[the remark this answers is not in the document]
angelojcruzjr: This is in as of https://github.com/UnitedIncome/components/pull/867/commits/13a29992ae9f227b4d94ca78a84cefb49ff437b0

angelojcruzjr:
@JamesIves @RohitRajendran made another round of updates that should address all the feedback up to this point 😇

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,59 @@
...
+    backgroundImage: `url(${imageURL})`,
+  };
+
+  const infoBoxClass = classNames({
+    'uic--info-box': true,
+    className,
+  });
+
+  return (
+    <aside className={infoBoxClass}>
```
RohitRajendran: Also allow passing in style

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,59 @@
...
+      <div
+        className="uic--info-box-title-area"
+        style={imageURL ? cardStyle : ''}
+      >
+        <h1 className="uic--info-box-title-text uic--header-2-text">{title}</h1>
```
RohitRajendran: I don't think h1 is right for this. There should only be one h1 on the page.

components/atoms/InfoBox/InfoBox.mdx
```diff
@@ -0,0 +1,60 @@
...
+InfoBox comes with several helper styles that make it easy for you to adhere to
+the style guide - especially when adding a footer. If you want to add a nicely aligned
+row such as what is seen in `InfoBoxFormattedRows`, add the following classes to the row parent element:
+`uic--info-box-formatted-row uic--d-flex uic--justify-content-between`. Labels can then be wrapped
+in the `uic--body-text` class while values can be wrapped in `uic--mcg-subhead-text`. Additionally,
+it should be noted that when you add footers we automatically create a light gray `hr` divider between
+the card content and the footer. Below is an example footer that would create a nicely formatted row:
+
+```javascript
+<div className="uic--info-box-formatted-row uic--d-flex uic--justify-content-between">
```
RohitRajendran: I think this is one of the situations that it's a lot better to have those flex properties be part of that custom class so that users don't have to add a bunch of classes to get the same styling

@JamesIves

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -0,0 +1,42 @@
+$min-height: 10rem;
+$background-size: 17rem;
+$container-min-height: 2.4rem;
+
+hr {
```
RohitRajendran: hr styling should be scoped inside of uic--info-box so that the styling doesn't bleed

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
```diff
@@ -0,0 +1,54 @@
...
+        >
+          <div className="uic--body-text">{row.label}</div>
+          <div className="uic--mcg-subhead-text">{row.value}</div>
+        </div>
+      ))}
+    </Fragment>
+  );
+};
+
+InfoBoxFormattedRows.propTypes = {
```
RohitRajendran: Need to also allow footer to be passed through

RohitRajendran: And style

components/atoms/InfoBox/InfoBox.mdx
angelojcruzjr: I agree - let's see if we can get consensus 😊

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Hmmm. Which would you say is appropriate here? `h2`?

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
angelojcruzjr: > Need to also allow footer to be passed through

So I currently have the `footer` be passed through the `InfoBox` shell component. Are you saying it should go in `InfoBoxFormattedRows`?

RohitRajendran: Ohhh jk, I was thinking of this wrong

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
```diff
@@ -0,0 +1,54 @@
+/** @module InfoBoxFormattedRows */
+import PropTypes from 'prop-types';
+import React, {Fragment} from 'react';
+import classNames from 'classnames';
+
+/**
+ * Renders an InfoBox with the title formatted the same way the
```
RohitRajendran: This needs to be updated

components/atoms/InfoBox/InfoBox.mdx
```diff
@@ -0,0 +1,60 @@
...
+  footer="Some Funky (Optional) Footer!"
+/>
+```
+
+## Styling Classes 👔
+
+InfoBox comes with several helper styles that make it easy for you to adhere to
+the style guide - especially when adding a footer. If you want to add a nicely aligned
+row such as what is seen in `InfoBoxFormattedRows`, add the following classes to the row parent element:
+`uic--info-box-formatted-row uic--d-flex uic--justify-content-between`. Labels can then be wrapped
```
RohitRajendran: I think this was actually more necessary/useful when InfoBoxFormattedRows was more of a wrapper component but since it's meant to be plugged into the content/footer. I don't think this is really necessary to call out anymore. Maybe we should change this section to talk about how you can use the InfoBoxFormattedRows with this

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.js
angelojcruzjr: Updated in https://github.com/UnitedIncome/components/pull/867/commits/529165bfc44344c47c80e8fc57ee90424707cc07

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Made `h2` in https://github.com/UnitedIncome/components/pull/867/commits/529165bfc44344c47c80e8fc57ee90424707cc07

components/atoms/InfoBox/InfoBox.mdx
angelojcruzjr: Updated in https://github.com/UnitedIncome/components/pull/867/commits/7697e6518948ab700c97b7d5bb7c6d6e4b6fdd90

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: Updated in https://github.com/UnitedIncome/components/pull/867/commits/7697e6518948ab700c97b7d5bb7c6d6e4b6fdd90

components/atoms/InfoBox/InfoBox.mdx
angelojcruzjr: Made this change in https://github.com/UnitedIncome/components/pull/867/commits/7697e6518948ab700c97b7d5bb7c6d6e4b6fdd90

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Allowed it here and in `InfoBoxFormattedRows` with changes in https://github.com/UnitedIncome/components/pull/867/commits/7697e6518948ab700c97b7d5bb7c6d6e4b6fdd90

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Renders a basic card with a title component on top and
+ * description/more information below the title portion.
+ * @param {Object} props Object containing the component props
+ * @returns {JSX} InfoBox JSX
+ */
+const InfoBox = (props) => {
+  const {title, content, imageURL, footer, className, style} = props;
+  let appliedStyle = {};
```
JamesIves: This is minor (and optional), but this could probably be cleaned up a little bit.

```
const appliedStyles = {...style}

if (imageURL) {
   appliedStyles.backgroundImage: `url(${imageURL})`
}
```

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,61 @@
...
+
+  if (imageURL) {
+    appliedStyle = {...style, backgroundImage: `url(${imageURL})`};
+  } else {
+    appliedStyle = {...style};
+  }
+
+  const infoBoxClass = classNames({
+    'uic--info-box': true,
+    className,
```
JamesIves: I think you need to pass `className` in as the second argument here.

Example:

```
      {
        'uic--itemization-box': true,
      },
      className,
    );
```

components/atoms/InfoBox/InfoBox.mdx
```diff
@@ -0,0 +1,52 @@
...
+While InfoBox gives you the flexibility to pass in any custom content and/or footer and have them be rendered
+to your speicifications, it also has built-in logic that allows you to create neatly formatted rows of
+`(label, value)` pairs. The [InfoBoxFormattedRows](/?path=/docs/molecules-infoboxformattedrows--default) component
+allows you to easily pass in your data as an array of JS objects and have it render neatly within the InfoBox component.
+Additionally, it should be noted that when you add footers we automatically create a light gray `hr` divider between
+the card content and the footer.
+
+## Props 🔧
+
+<Props of={InfoBox} />
```
JamesIves: In some of the components we have a `Related` header that links to similar things. Can we add that here? 

Here's the template:

```suggestion
<Props of={InfoBox} />

## Related ✨

- [Checkboxes](/?path=/story/molecules-checkboxes--default)
```

We should link to InfoBoxFormatted and do the same in the other readme.

components/atoms/InfoBox/InfoBox.mdx
JamesIves: I agree in this case

components/molecules/InfoBoxFormattedRows/InfoBoxFormattedRows.mdx
```diff
@@ -0,0 +1,78 @@
+import {Preview, Props} from '@storybook/addon-docs/blocks';
```
RohitRajendran: I don't think this component should be a molecule anymore. Seems more like an atom

angelojcruzjr: 😱I'll move it over before merging.

JamesIves: I'm looking forward to re-organizing Storybook, most of the library definitions are way off 😞

angelojcruzjr:
@RohitRajendran made changes that allow for the InfoBox to have the same functionality as what is currently in the app. Also moved it over to Atoms and added tests/a story for a small utility function that I created. I'll merge this once you give your blessing 🙏🏽

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,161 @@
+/* eslint-disable jsx-a11y/mouse-events-have-key-events */
+/** @module InfoBox */
+import PropTypes from 'prop-types';
+import React, {Component} from 'react';
+import './InfoBox.scss';
+import classNames from 'classnames';
+import {formatEllipsis} from '../../utilities/FormatUtils/FormatUtils';
+
+const CHAR_COUNT = 25;
```
JamesIves: Can we make this a prop and just set this as the default value?

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,161 @@
...
+  /** Optional footer JSX to be displayed below the content. */
+  footer: PropTypes.node,
+  /** Optionally make the InfoBox card appear clickable/selectable. */
+  selectable: PropTypes.bool,
+  /** Additional class names to apply to the container. */
+  className: PropTypes.string,
+  /** Additional style properties to apply to the container. */
+  style: PropTypes.objectOf(PropTypes.string),
+  /** Optional onClick handler. */
+  onClick: PropTypes.func,
```
JamesIves: This doesn't appear to be optional;

```
  handleClick(event) {
    event.stopPropagation();

    return this.props.onClick();
  }
```

If `onClick` isn't provided it's going to error. We should check its existence before firing it. Either that or mark thisprop has required.

components/utilities/FormatUtils/FormatUtils.mdx
```diff
@@ -41,3 +41,15 @@ FormatUtils.formatCurrencyNoDecimal(2000);
...
 ```
+
+### formatEllipsis
+
+Formats a string into a truncated version based on a specified character limit.
+
+```javascript
+import {FormatUtils} from '@unitedincome/components';
+
+FormatUtils.formatEllipsis('This is a very, very long string', 25);
```
JamesIves: Is it ok for the title to wrap on small screen sizes?

Screen Shot 2019-11-25 at 12 51 03 PM

I typically like to use CSS for things like this because it allows you specify a specific width percentage that should display prior to the ellipsis generation, as opposed to a hard character limit. Both methods have their drawbacks though.

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,161 @@
...
+};
+
+InfoBox.defaultProps = {
+  selectable: false,
```
JamesIves: Does this need to be a default prop? Isn't `selectable` just going to default to falsey?

angelojcruzjr: That's true. Got rid of it in https://github.com/UnitedIncome/components/pull/867/commits/49076394869a5a2bd9eedb386d56ec73139ae885

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Updated to check for existence before firing in latest push (https://github.com/UnitedIncome/components/pull/867/commits/49076394869a5a2bd9eedb386d56ec73139ae885)

components/atoms/InfoBox/InfoBox.js
angelojcruzjr: Yeah that's a lot better. Went ahead and made this an optional prop in https://github.com/UnitedIncome/components/pull/867/commits/49076394869a5a2bd9eedb386d56ec73139ae885

components/utilities/FormatUtils/FormatUtils.mdx
angelojcruzjr: So I went back and forth on how this should look and after talking w/ Monique @RohitRajendran and I decided to just use the functionality that is currently in place of the app which is using a hard character limit. I have a short video of the current functionality that I can send you if you're curious (github won't let me upload .mov's 😞) but I can confirm this is how it currently looks. This will probably change in the future but for now we are just keeping it status quo, so to speak.

JamesIves: Sounds good to me

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,164 @@
...
+
+    return (
+      <aside {...containerProps}>
+        <div
+          className="uic--info-box-title-area"
+          style={imgStyle}
+          onMouseEnter={this.handleMouseOver}
+          onMouseLeave={this.handleMouseOut}
+        >
+          <h2 className="uic--info-box-title-text uic--header-2-text">
```
RohitRajendran: If you look at the comps, I don't think h2 makes sense from an information hierarchy standpoint.

RohitRajendran: Since it can depend on the page that it's one, I think we should skip using h tags for this.

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,164 @@
...
+  /** Optional number of characters to limit the title to before styling it as an ellipsis. */
+  titleCharLimit: PropTypes.number,
+  /** The content JSX to be displayed below the title. */
+  content: PropTypes.node.isRequired,
+  /** Optional image URL string to place in your InfoBox - we use the URL to create an image that appears to the right of the title. */
+  imageURL: PropTypes.string,
+  /** Optional footer JSX to be displayed below the content. */
+  footer: PropTypes.node,
+  /** Optionally make the InfoBox card appear clickable/selectable. */
+  selectable: PropTypes.bool,
```
RohitRajendran: Can't you just infer this from if onClick was passed through?

components/atoms/InfoBox/InfoBox.js
```diff
@@ -0,0 +1,164 @@
...
+      className: containerClass,
+      style,
+    };
+
+    if (imageURL) {
+      imgStyle.backgroundImage = `url(${imageURL})`;
+    }
+
+    if (onClick) {
+      containerProps.onClick = this.handleClick;
```
RohitRajendran: Can't you just pass `this.props.onClick` directly instead of using a wrapper function?

JamesIves: @angelojcruzjr Can we address these in a separate PR?

angelojcruzjr: Si. Will create a new one for @RohitRajendran 's comments

PR #868, opened 2019-11-20
HouseIcon Component

JamesIves:
Shifts the house icon component from the Dashboard into the component library. You can define `x` amount of houses using the `houses` prop, and you can highlight `x` amount of them using the `highlight` prop. You're able to display a single house if that's preferred too. 

* Take a look at the component in the library!

You can't customize the fill due to the nature of the component. 


components/atoms/icons/HouseIcon/HouseIcon.js
```diff
@@ -0,0 +1,117 @@
...
+        style={{
+          width,
+          height,
+          ...style,
+        }}
+        key={i}
+      >
+        <svg viewBox="0 0 30 30" preserveAspectRatio="none">
+          <g stroke="none" strokeWidth="1" fill="none" fillRule="evenodd">
+            {(item < highlight - 1 || item === highlight - 1) && !isIE ? (
```
RohitRajendran:
```
            {(item < highlight - 1 || item === highlight - 1) && !isIE && (
```
You could just use `&&` instead of a ternary

JamesIves: Good catch

components/atoms/icons/HouseIcon/HouseIcon.js
```diff
@@ -0,0 +1,117 @@
...
+  return [...Array(houses).keys()].map((item, i) => {
+    return (
+      <div
+        className={containerClasses}
+        style={{
+          width,
+          height,
+          ...style,
+        }}
+        key={i}
```
angelojcruzjr: React [doesn't like when you use indices as keys](https://reactjs.org/docs/lists-and-keys.html#keys). Do we have a utility to create keys that we can use in place of this? I see that in pretty much every other place we also use an index as a key (example: `ItemizationWidget.js` line 193). Maybe this could become a story to replace all these with a utility?

JamesIves: For a quick win I'll combine the id with the index to create the key. Creating a utility may be a good idea but probably not the most pressing thing in the world. Let's create a story for it.

RohitRajendran: I don't think combining it with id is really doing anything better. The main reason why they say not to use index with keys is still there.

PR #870, opened 2019-11-22
[APP-3257] ItemizationBox Component

JamesIves:
Implements the `<ItemizationBox />` component. This component itemizes a list of values/labels and creates a total at the top. You can collapse the component to reveal each itemization, and also include an optional button and tooltip. Additionally if the value goes into the negatives it will be reflected with `$pink`.

Screen Shot 2019-11-22 at 11 20 08 AM
Screen Shot 2019-11-22 at 11 20 22 AM
Screen Shot 2019-11-22 at 11 20 30 AM

* Load up the component library and look at the `ItemizationBox` component.
* Ensure the values line up, and that it matches the comp.

I will verify with Monique to make sure everything looks good with her before merging.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-3257)


components/atoms/icons/HouseIcon/HouseIcon.mdx
```diff
@@ -36,7 +36,7 @@ You can animate the fill on the houses by providing a `highlight` prop to the co
 </Preview>
 
 ```javascript
-<HouseIcon height="3rem" width="3rem" houses={10} highlight={50} />
+<HouseIcon height="3rem" width="3rem" houses={100} highlight={50} />
```
JamesIves: This is unrelated, just something I missed in the last PR for the `HouseIcon` component.

components/atoms/ItemizationBox/ItemizationBox.js
```diff
@@ -0,0 +1,201 @@
...
+                  return (
+                    <div className="uic--row" key={index}>
+                      {item.tooltip ? (
+                        <div className="uic--col-8">
+                          <div className="uic--row uic--align-items-center">
+                            <div className="uic--itemization-box__itemizations-title uic--col-sm-6 uic--col-5">
+                              {item.label}
+                            </div>
+                            <div className="uic--itemization-box__itemizations-tooltip__label uic--d-flex uic--col-sm-6 uic--col-7">
+                              (
```
matthewheck: what's with the ()?

JamesIves: The tooltip label gets wrapped in brackets in the comps so I've just hardcoded it here. This way you can put `label: 'cat facts` in the payload. We could get rid of it and make the user place `label: '(cat facts)'` in the payload but then consistency isn't enforced.

components/atoms/ItemizationBox/ItemizationBox.js
```diff
@@ -0,0 +1,201 @@
...
+
+    // Toggles the height if the component was previously collapsed.
+    if (prevProps.isCollapsed !== this.props.isCollapsed) {
+      updatedState.height = this.props.isCollapsed
+        ? 0
+        : this.contentNode.current.scrollHeight;
+    }
+
+    if (Object.keys(updatedState).length > 0) {
+      this.setState(updatedState);
```
RohitRajendran: Couldn't you just put this in with the previous if statement

components/atoms/ItemizationBox/ItemizationBox.js
```diff
@@ -0,0 +1,201 @@
...
+    }
+  }
+
+  /** Sums the 'items' and 'value' properties to form a group total.
+   * @param {array} values - The values to itemize.
+   * @returns {integer} - Returns the sum.
+   */
+  sumTotal(values) {
+    return values.reduce((acc, obj) => {
+      let accumulator = acc;
```
RohitRajendran: What's the point of this? This is the same as just modifying acc directly

components/atoms/ItemizationBox/ItemizationBox.js
```diff
@@ -0,0 +1,201 @@
...
+      accumulator += obj.value || 0;
+      accumulator += obj.items
+        ? obj.items.reduce((a, o) => {
+            return 'value' in o ? o.value + a : a;
+          }, 0)
```
RohitRajendran:
```suggestion
          }, obj.value || 0)
```
Couldn't you simplify with doing something like this?

components/atoms/ItemizationBox/ItemizationBox.js
RohitRajendran: Would it ever be an icon for the label?

components/atoms/ItemizationBox/ItemizationBox.js
```diff
@@ -0,0 +1,201 @@
...
+              </div>
+            </animated.div>
+          )}
+        </Spring>
+      </div>
+    );
+  }
+}
+
+ItemizationBox.propTypes = {
```
RohitRajendran: Does this need to support an onclick handler to support click to collapse/uncollapse

JamesIves: It does not

components/atoms/ItemizationBox/ItemizationBox.js
JamesIves: Yeah, that's probably better.

components/atoms/ItemizationBox/ItemizationBox.js
JamesIves: Text only apparently

components/atoms/ItemizationBox/ItemizationBox.js
JamesIves: Not sure, this is left over code from something else. I removed it.

components/atoms/ItemizationBox/ItemizationBox.js
JamesIves: This actually doesn't need to be here at all so I'm going to remove it, I borrowed some of this function from the other itemization component, and `items` doesn't exist as a concept here.

PR #891, opened 2019-12-04
useState Hook Example With Tests

JamesIves:
This is an exmaple of the useState hook with unit tests. I'd like to discuss implementing more of these in the component library so long as the test coverage is sufficient enough.

Right now you can run these unit tests directly using `yarn testSingle`, for whatever reason when you run the entire test suite it doesn't work correctly. This would need to be resolved before we move forward with any of these. 

* The component should behave as intended.
* The unit tests should run. Once the full test suite builds the test coverage should remain the same.

Thanks to @labichn for insight on the workings of this. I'd love to figure out a way around the `require` usage. It seems that [Jest handles mocking hooks better](https://itnext.io/testing-components-built-using-react-hooks-with-jest-enzyme-edb87d703756), perhaps switching to using it in this library may be a good idea. Innovation sprint idea? 


labichn:
I'd be hesitant to start incorporating hooks into our components until the Suspense API and Concurrent mode are stable (as Dan Abramov has suggested: https://twitter.com/dan_abramov/status/1057011888335347713?lang=en). It's a significant change, and I don't think there are any obvious immediate benefits.

I've been building a side-project using Concurrent mode and Suspense, but from my few experiments those features are far from production ready.

JamesIves:
> I'd be hesitant to start incorporating hooks into our components until the Suspense API and Concurrent mode are stable (as Dan Abramov has suggested: https://twitter.com/dan_abramov/status/1057011888335347713?lang=en). This way we can rethink our state management as a whole, rather than just doing some new stuff with hooks. It's a significant change, and I don't think there are any obvious immediate benefits.
> 
> I've been building a side-project using Concurrent mode and Suspense, but from my few experiments those features are far from production ready.

Is this really necessary for this component library though? I feel like these components are low level enough where we can safely start to implement hooks without seeing side effects elsewhere.

matthewheck:
I think it's fair to see that we shouldn't adopt a change like this unless we have some type of plan for how we can benefit from it. That said, I think @JamesIves is demonstrating potential usage here in a productive way, even if this PR probably shouldn't be merged.

I'm not convinced that a massive back-and-forth on a GitHub PR is a productive use of time. I'd much rather we have a conversation as a team about this. Shall we schedule a time sometime?

JamesIves:
To be clear this PR **shouldn't** be merged. I'd love to put something and hold a meeting internally about this.

PR #896, opened 2019-12-05
[APP-4148] Re-order card errors so the most urgent appears first

CalMlynarczyk:
### Description
Monique, Alicia and I talked about how to handle multiple error messages at the bottom of cards. We decided to re-order the error messages so at least the most "important" one was on top and additional error messages display below.

### Testing Instructions
1. Go to an input card; leave the value blank
2. Turn on `clearFuture` and `hasMadeChanges` and turn off `isLatestCard`
3. Click _Continue_ three times to make the "required fields" message also appear

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4148)


components/organisms-simple/QuestionCard/QuestionCard.js
```diff
@@ -63,13 +63,14 @@ const QuestionCard = ({
       }
       beforeButton={
         <Fragment>
-          {beforeButton}
           {hasMadeChanges && showEditWarning && (
             <p className="uic--warning-message">
-              Note: Changing your answer to this question will require you to
-              answer additional questions.
+              Note: The answer to this question has an impact on subsequent
```
RohitRajendran: Don't think this copy should be getting changed

CalMlynarczyk: I didn't catch that after I rebased.

PR #920, opened 2019-12-10
forwardRef for Button

JamesIves:
Allows ref forwarding for the `Button` component.

* Nothing should change. However you should now be able to add `ref={React.createRef}` to forward a ref onto the button container.

N/A

N/A


JamesIves:
@RohitRajendran I think it will break tests that reference `'Button'` as opposed to `Button`. Functionality wise though I don't think so mainly because we don't forward refs right now. 

However I'd like to make this change as part of a larger release and do the major version bump as a pre-caution.

PR #932, opened 2019-12-16
InVision DSM Integration

JamesIves:
Added the InVision implementation and added all available InVision id's to the stories. 🎉

I haven't added the McG ones as there's some DSM organization that needs to occur first. 

* Add the required `.dsmrc` file to your local configuration (it is git ignored on purpose) and run `yarn dsm-storybook:publish`.
* I've updated the secrets menu with the required settings. The GitHub action workflow will run and generate the `.dsmrc` file, publish it, and then cleanup the files so it doesn't somehow get accidentally committed. 

InVision can only store a single id, therefore the default story in most cases gets applied with the id.


matthewheck:
Those are all hard-coded IDs, right? What's the process for generating them and making sure they're correct and such?

JamesIves:
@matthewheck These are hard coded yes, in the InVision app you click the "React" button in the component screen and it generates an id. This just allows InVision to know where to publish the story. 

Screen Shot 2020-01-22 at 10 23 06 AM

.github/workflows/deployment-prod.yml
```diff
@@ -22,6 +22,22 @@ jobs:
       - name: Install
         run: yarn install
 
+      - name: Install InVision .dsmrc
+        run: |
+          echo > '{
```
JamesIves: I'm aware this is weird. This data _has_ to source from a `.dsmrc` file and not an environment variable according to InVision.

PR #939, opened 2019-12-20
Deployment Changes

JamesIves:
Removes the AWS deployment job so it's now hosted on GitHub pages.

* Once this pushes to dev we will see it build there. I may need to make a couple of minor changes.
* This does not replace the entire CircleCI job. It probably doesn't really need to, but it may be better to not have two CI tools.


.circleci/config.yml
```diff
@@ -62,17 +62,6 @@ jobs:
...
-          command: |
-            sudo apt install python-dev python-pip
-            sudo pip install awscli
-
-      - run:
-          name: Publish Storybook
-          command: yarn publishStorybook
-
       - run:
           name: Run IE11 Smoke Test
```
CalMlynarczyk: What is the new URL going to be? We need to update the URL in the Nightwatch config if it is changing since the old domain won't do anything anymore.

JamesIves: `https://unitedincome.github.io/components/` for prod, `https://unitedincome.github.io/components/develop` for dev.

JamesIves:
Waiting on @dwolfand to apply an access token and we're good to go here.

PR #961, opened 2020-01-03
QuestionAnswer Component - header can take node + string

ortizlu:
Add ability to pass in a node in the header of the QuestionAnswer component.

Pass in a node in the header and make sure everything looks okay.


components/molecules/QuestionAnswer/QuestionAnswer.js
```diff
@@ -46,9 +46,13 @@ class QuestionAnswer extends PureComponent {
       <div className={containerClasses} style={style}>
         <div className="uic--row">
           <div className="uic--question-answer__header uic--d-flex uic--justify-content-center uic--align-items-center uic--col-12 uic--col-md-6">
-            <h2 className="uic--question-answer__header-inner uic--header-1-5-text uic--text-center">
+            {typeof header === 'string' ? (
+              <h2 className="uic--question-answer__header-inner uic--header-1-5-text uic--text-center">
```
JamesIves: We should maintain some form of consistency here, perhaps wrapping this block in another div that holds the `uic--question-answer__header-inner` class. Either that or add a div on the node variant that also has this class.

These classes are equal parts about styling the elements and making it so it's easier to navigate through the DOM to find a particular element in a component.

components/molecules/QuestionAnswer/QuestionAnswer.js
```diff
@@ -83,7 +87,7 @@ QuestionAnswer.propTypes = {
   /** Additional style properties to apply to the container. */
   style: PropTypes.objectOf(PropTypes.string),
   /** The header of the QuestionAnswer section. */
-  header: PropTypes.string.isRequired,
+  header: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
```
JamesIves: We should add an example of this in the readme with some additional styling context. If you take a look at the `Cabinet` component we give some additional instructions on how to maintain styles when adding markup via available class names.

RohitRajendran:
```suggestion
  header: PropTypes.node.isRequired,
```
Don't think you actually need to specify both string and node

components/molecules/QuestionAnswer/QuestionAnswer.js
RohitRajendran: @ortizlu What's the use case for this change? Is it just to be able to do bold/italics in the header or were there additional changes that needed to be supported?

JamesIves: It's so you can add a paragraph into the block below the header.

components/molecules/QuestionAnswer/QuestionAnswer.js
```diff
@@ -82,8 +85,10 @@ QuestionAnswer.propTypes = {
   className: PropTypes.string,
   /** Additional style properties to apply to the container. */
   style: PropTypes.objectOf(PropTypes.string),
-  /** The header of the QuestionAnswer section. */
-  header: PropTypes.string.isRequired,
+  /** The title of the QuestionAnswer section. */
+  title: PropTypes.string,
```
JamesIves: We probably shouldn't change the prop name for the header as it's technically a breaking change and would require a larger version bump. I'd prefer to name these `header` and `description` to keep them consistent. Technically speaking as this project is open source we should be making changes as-if other projects outside of United Income also use the same components

components/molecules/QuestionAnswer/QuestionAnswer.js
```diff
@@ -46,9 +46,12 @@ class QuestionAnswer extends PureComponent {
...
         <div className="uic--row">
           <div className="uic--question-answer__header uic--d-flex uic--justify-content-center uic--align-items-center uic--col-12 uic--col-md-6">
-            <h2 className="uic--question-answer__header-inner uic--header-1-5-text uic--text-center">
-              {header}
-            </h2>
+            <div className="uic--question-answer__header-inner">
+              <h2 className="uic--question-answer__header-inner-title uic--header-1-5-text uic--text-center">
+                {title}
+              </h2>
+              {subtitle}
```
RohitRajendran: We should set up the subtitle similar to how we setup description in then Question Card, conditionally wrapping it in a p if a string https://github.com/UnitedIncome/components/blob/72733ea61bac9666172a1ebb82213d3db85e81f1/components/organisms-simple/QuestionCard/QuestionCard.js#L47

components/molecules/QuestionAnswer/QuestionAnswer.js
ortizlu: Good point! Changed back to header. Does the new CSS/HTML make sense?

components/molecules/QuestionAnswer/QuestionAnswer.js
```diff
@@ -82,8 +94,10 @@ QuestionAnswer.propTypes = {
   className: PropTypes.string,
   /** Additional style properties to apply to the container. */
   style: PropTypes.objectOf(PropTypes.string),
-  /** The header of the QuestionAnswer section. */
-  header: PropTypes.string.isRequired,
+  /** The main header of the QuestionAnswer section. */
+  header: PropTypes.string,
```
RohitRajendran: Is removing the `isRequired` intentional?

components/molecules/QuestionAnswer/QuestionAnswer.js
```diff
@@ -33,7 +33,16 @@ class QuestionAnswer extends PureComponent {
 
   /** @inheritdoc */
   render() {
-    const {className, style, items, header} = this.props;
+    const {className, style, items, header, description} = this.props;
+
+    const descriptionNode =
+      typeof description === 'string' ? (
+        <p className="uic--question-answer__header-inner-subtitle">
```
RohitRajendran: Is this class actually necessary?

components/molecules/QuestionAnswer/QuestionAnswer.spec.js
```diff
@@ -5,7 +5,7 @@ import QuestionAnswer from './QuestionAnswer';
 
 test('QuestionAnswer - renders', (t) => {
   const props = {
-    header: 'FAQs',
+    title: 'FAQs',
```
RohitRajendran: Shouldn't these be `header`?

components/molecules/QuestionAnswer/QuestionAnswer.spec.js
```diff
@@ -26,7 +26,8 @@ test('QuestionAnswer - renders', (t) => {
 
 test('QuestionAnswer - setExpanded', (t) => {
   const props = {
-    header: 'FAQs',
+    title: 'FAQs',
+    subtitle: <p>This is subtitle.</p>,
```
RohitRajendran: Shouldn't this be `description`?

components/molecules/QuestionAnswer/QuestionAnswer.js
ortizlu: it was not.

components/molecules/QuestionAnswer/QuestionAnswer.js
ortizlu: no I just wasn't sure if you needed every element to have markup for consistency

RohitRajendran: Let's wrap it and put the class on there then so that it still gets included when `description` is a node

ortizlu: just wrapped it

components/molecules/QuestionAnswer/QuestionAnswer.js
```diff
@@ -53,7 +53,9 @@ class QuestionAnswer extends PureComponent {
               <h2 className="uic--question-answer__header-inner-title uic--header-1-5-text uic--text-center">
                 {header}
               </h2>
-              {descriptionNode}
+              <div className="uic--question-answer__header-inner-subtitle">
```
RohitRajendran: `{descriptionNode && <div ....}`
No need to show the wrapper div if there's no description node

ortizlu: good catch, thank you! Just added that.

JamesIves:
@ortizlu I'm merging this so it goes into the next release.

PR #971, opened 2020-01-08
McGonagall Docs Re-Org / MVP Generator

JamesIves:
Improved the McGonagall Documentation by re-organizing it and providing some additional context. Also added a `yarn mcgonagall` command that allows you to generate a new flow within the component library, this allows us to keep the docs consistent in an open source environment even though our internal generator will exist within the client repo.

I've also fixed the example for the xState visualizer so you can just copy and paste it straight in to see it working. 

* Run the `yarn mcgonagall` command and create a flow.
* Read the docs by going to `Templates/McGonagall` and click `Docs`.
* Paste the generated flow into the client and fix the imports.
* You should have a working flow and the docs should be enough to get you up and running.

I'll move the same generator into client if this looks good to everyone. There's less todo items in this version so it's quite speedy to get it up and running.

Give this a read and let me know if you have a good understanding as to what's going on with McGonagall. The idea is to let new people read this documentation so they can start working with it right away.


components/templates/McGonagall/McGonagall.mdx
```diff
@@ -19,49 +19,111 @@ import {McGonagall} from '@unitedincome/components';
 
 ## Example 🚀
 
-There's 4 key parts to creating a flow in McGonagall: `stateConfig`, `stateOptions`, `stateContext`, and `steps`. Depending on the project a scaffolder tool may be available to you by running `yarn generate mcg new`.
+You can generate a new McGonagall flow by running the following command within this library and answering a few basic questions. Although not neccersary for all implementations, the generateor will assumes you're using Redux in your project.
```
labichn: s/neccersary/necessary/

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -19,49 +19,111 @@ import {McGonagall} from '@unitedincome/components';
...
 
-1. In the `.xstate.js` file, lay out the flow by adding state nodes to the `stateConfig` with the names for the conditions
-2. In the `.steps.js` file, add an object to the steps array for each state node that you created in the `stateConfig`
-3. In the `.js` file, fill out the `stateContext` object with any data you need to pull in
-4. Starting with the first step, start adding the render card function for the step in `.step,js` and updating the `stateConfig` and the `stateOptions` with necessary actions and guards
+```bash
```
labichn: Nit: I'd stick with `sh` over `bash` unless you're using `bash`-specific features.

labichn:
@RohitRajendran, @JamesIves, is [this](https://github.com/UnitedIncome/components/pull/971/files#diff-fc9e1ee0bf78a45f71d8c559b9660c68L290) description of a Card's `props.context` accurate? Can we clarify how it's different than the [state](https://github.com/UnitedIncome/components/pull/971/files#diff-fc9e1ee0bf78a45f71d8c559b9660c68L306) passed into the card? By these two descriptions, they seem to be the same thing, but I don't think that's the case.

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -270,17 +315,17 @@ const steps = [
...
 ];
 ```
 
-#### Card
+#### steps.card
 
 The function in the card property will get passed 2 arguments, props and state.
 
-The props argument will include information from the mcgonagall framework and was designed so you can pass it straight through to the cards:
+The props argument will include information from the Mcgonagall framework and was designed so you can pass it straight through to the cards:
```
labichn: s/Mcgonagall/McGonagall/

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -348,35 +393,9 @@ const steps = [
...
-
-The state machine's context is replicated in the local component state. This is so that you can make changes in a card, decide to cancel it, and revert back to the original value. The local component state values get changed when the `onChange` function gets called. When the `onSubmit` function is called, it pulls the outputs you specified in the `steps` object from the component state and passes it through as the event payload of the state machine.
-
-## Visualizing the Flow
+## Visualizing the Flow 💡
 
-Thanks to xState, it's now much easier to visualize the entire flow. You just have to copy and paste the `stateConfig` and `stateOptions` (which is essentially the entire `.xstate.js` file except for the import and exports) into the [xState visualizer tool](https://statecharts.github.io/xstate-viz/) and pass both of the variables into the machine like so:
+If you'd like to visualize your McGonnagal flow you can copy and paste the `stateConfig` and `stateOptions` objects into the [xState visualizer tool](https://statecharts.github.io/xstate-viz/) and pass both of the variables into `Machine` like so.
```
labichn: s/McGonnagal/McGonagall/

_templates/mcgonagall/new/cards.ejs.t
```diff
@@ -0,0 +1,61 @@
...
+---
+<%
+  compName = h.changeCase.pascal(name)
+-%>
+import {
+  CompletionScreen,
+  MessageCard,
+} from '@unitedincome/components';
+import React from 'react';
+import {outputs} from './<%= compName %>.constants';
```
matthewheck: `outputs` isn't used anywhere in here. I think it'd be great to add an example of what that might look like in practice.

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -419,19 +438,19 @@ const stateOptions = {
...
+    hasNameAndIsMarried: (ctx, eve) => true,
+    hasNameAndIsNotMarried: (ctx, eve) => true,
   },
 };
 
 const fetchMachine = Machine(stateConfig, stateOptions);
 ```
 
-## Utilities
+This will produce the following, allowing you to see how each question transitions and affects eachother.
```
labichn: s/eachother/each other/

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -443,6 +462,10 @@ The `saveToStateContext` function will allow you to save data to your state cont
 import {saveToStateContext} from '@unitedincome/components/dist/mcgonagall';
 ```
 
+### Typography Styles
+
+There are a number of typograhy utility classes that be found [here](/?path=/story/utilities-styles--typography).
```
labichn: s/typograhy/typography/

JamesIves:
@labichn Can you flag the line? I can clarify but I'm not sure where exactly that is

labichn:
> @labichn Can you flag the line? I can clarify but I'm not sure where exactly that is

Sorry! The lines are in the links, but I guess they don't work: 290 and 306 from McGonagall.mdx.

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -3,7 +3,7 @@ import McGonagall from './McGonagall';
 
 # McGonagall
 
-McGonagall is a question flow framework. McGonagall uses [xState](https://xstate.js.org/docs/), a finite-state machine library, to help it navigate through the flow.
+McGonagall is a question flow framework which uses [xState](https://xstate.js.org/docs/). It's used to generate a large set of data in the form of a questionaire.
```
matthewheck:
```suggestion
McGonagall is a question flow framework that uses [xState](https://xstate.js.org/docs/). It's used to generate a large set of data in the form of a questionnaire.
```

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -19,49 +19,111 @@ import {McGonagall} from '@unitedincome/components';
...
+/>
+```
 
-  // Initial state node
-  initial: 'first',
+### stateContext 💡
+
+The `stateContext` prop is used to setup the initial McGonagall state. You can use this prop to pass in methods or pieces of data that are sourced from outside of the question flow.
+
+In the following example `futureOrCurrent` is not asked in the flow, but the information is used to influence questions within the flow. With this setup anytime you have access to McGonagall's state you'll be able to reference `state.futureOrCurrent`. This can be useful if you need to pass in personalized information to reference within a question, for example a users name or something similar.
```
matthewheck:
```suggestion
In the following example `futureOrCurrent` is not asked in the flow, but the information is used to influence questions within the flow. With this setup anytime you have access to McGonagall's state you'll be able to reference `state.futureOrCurrent`. This can be useful if you need to pass in personalized information to reference within a question, for example a user's name or something similar.
```

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -19,49 +19,111 @@ import {McGonagall} from '@unitedincome/components';
...
+      />
+    );
+  }
+}
+```
+
+### stateConfig 💡
+
+The `stateConfig` prop is where you'll configure the transitions between each step in the question flow, this should be imported from the `Flow.xstate.js` file. Here you can add logic to determine which question is presented after the current one, and logic that
+ensures that certain information is present before the user will be allowed to advance.
```
matthewheck: weird line break

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -19,49 +19,111 @@ import {McGonagall} from '@unitedincome/components';
...
+        isFetching={this.props.isFetching}
+      />
+    );
+  }
+}
+```
+
+### stateConfig 💡
+
+The `stateConfig` prop is where you'll configure the transitions between each step in the question flow, this should be imported from the `Flow.xstate.js` file. Here you can add logic to determine which question is presented after the current one, and logic that
```
matthewheck:
```suggestion
The `stateConfig` prop is where you'll configure the transitions between each step in the question flow, this should be imported from the `Flow.xstate.js` file. Here you can add logic to determine which question is presented after the current one and logic that
```

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -19,49 +19,111 @@ import {McGonagall} from '@unitedincome/components';
...
+  stateOptions={stateOptions}
+  steps={steps}
+/>
+```
 
-  // Initial state node
-  initial: 'first',
+### stateContext 💡
+
+The `stateContext` prop is used to setup the initial McGonagall state. You can use this prop to pass in methods or pieces of data that are sourced from outside of the question flow.
```
matthewheck:
```suggestion
The `stateContext` prop is used to setup the initial McGonagall state. You can use this prop to pass in functions or pieces of data that are sourced from outside of the question flow.
```

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -151,9 +212,7 @@ const stateOptions = {
...
 ```
 
-In an event for a state node, you can pass in an array of strings that correspond to the name of the action to excute.
-
-In the below example, there are 2 potential state nodes a user can go from the `name` state node. The state machine will first check if it passes the condition for the first object in the array so if the `hasNameAndIsMarried` guard returns true, the flow will transition to the owner step and run the 5 actions that are listed. If `hasNameAndIsMarried` returns false, it will then check the next condition. A state node should be able to pass at least one of the conditions.
+In an event for a state node, you can pass in an array of strings that correspond to the name of the action to excute. In the below example, there are 2 potential state nodes a user can go from the `name` state node. The state machine will first check if it passes the condition for the first object in the array so if the `hasNameAndIsMarried` guard returns true, the flow will transition to the owner step and run the 5 actions that are listed. If `hasNameAndIsMarried` returns false, it will then check the next condition. A state node should be able to pass at least one of the conditions.
```
matthewheck:
```suggestion
In an event for a state node, you can pass in an array of strings that correspond to the name of the action to execute. In the below example, there are 2 potential state nodes a user can go to from the `name` state node. The state machine will first check if it passes the condition for the first object in the array; so if the `hasNameAndIsMarried` guard returns true, the flow will transition to the owner step and run the 5 actions that are listed. If `hasNameAndIsMarried` returns false, it will then check the next condition. A state node should be able to pass at least one of the conditions.
```

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -219,27 +278,13 @@ name: {
...
-```javascript
-const stateContext = {
-  id: this.assetId,
-  ...asset,
-  userData: this.props.userData || null,
-  insertAssetValue: this.insertAssetValue,
-  saveUserData: this.props.saveUserData,
-};
-```
+When the McGonagall component first mounts, the framework uses the guards to rehydrate the flow. It tries transitioning at each step until it hits a step that it can't navigate away from because it doesn't satisfy any condition. It's important to add guards to each State Node possible transition so that it the flow starts back up at the appropriate place.
```
matthewheck:
```suggestion
When the McGonagall component first mounts, the framework uses the guards to rehydrate the flow. It tries transitioning at each step until it hits a step that it can't navigate away from because it doesn't satisfy any condition. It's important to add guards to each State Node possible transition so that the flow starts back up at the appropriate place.
```

_templates/mcgonagall/new/cards.ejs.t
JamesIves: I'm going to beef the boilerplate up with a couple of basic question types, stay tuned

JamesIves:
@labichn @matthewheck All of the feedback should be addressed now.

PR #973, opened 2020-01-09
McGonagall Doc Changes v2

JamesIves:
Adds some more context/docs for some common things.


components/templates/McGonagall/McGonagall.mdx
```diff
@@ -384,12 +384,173 @@ const steps = [
...
+  whose: 'whose',
+};
+
+const outputDefaults = {
+  whose: whoseEnum.MINE,
+};
+
+export const renderOwner = (props, state) => {
+  return (
+    <QuestionCard
```
RohitRajendran: Radio button card without summary would keep the example a little simpler to read

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -384,12 +384,173 @@ const steps = [
...
+      {...props}
+      title={`Will there be any additional costs associated?`}
```
RohitRajendran: This is unnecessary

components/templates/McGonagall/McGonagall.mdx
JamesIves: Changed

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -384,12 +384,173 @@ const steps = [
...
+    title: 'How will you acquire this?',
+    shortTitle: 'Purchased or Inherited',
+    outputs: [outputs.assetPurchased],
+    card: renderAcquisition,
+    clearFuture: true,
+  },
+];
+```
+
+This will clear the questions from the flow, but you'll still need to use an action to clear the value from McGonagall's state. In the following example `clearPurchaseDependencies` is used to clear any outputs generated by the `name` state node whenever this question has been changed. This will reset the values back to `null` so your state remains clean.
```
RohitRajendran:
```suggestion
This will clear the questions from the flow, but you'll still need to use an action to clear the value from McGonagall's state. In the following example `clearPurchaseDependencies` is used to clear any outputs generated by the `name` state node whenever this question has been changed. This will reset the values back to `undefined` so your state remains clean.
```

RohitRajendran: It might be good to differentiate how you're clearing it from the state context through `assign` and returning the value and with redux by using the `insertUserDataValue`

JamesIves: There's already a section on `assign` - I've made some changes to that paragraph to clarify the requirement of clearing both the redux store and the mcg state.

components/templates/McGonagall/McGonagall.mdx
```diff
@@ -384,12 +384,169 @@ const steps = [
...
+  whose: 'whose',
+};
+
+const outputDefaults = {
+  whose: whoseEnum.MINE,
+};
+
+export const renderOwner = (props, state) => {
+  return (
+    <RadioButtons
```
RohitRajendran:
```suggestion
    <RadioButtonCard
```

PR #978, opened 2020-01-10
Slider tooltip sticky fix 

RohitRajendran:
### Description
Fix slider when tooltip sticky variant is given but not tooltip sticky position

1. Go to the slider story
1. Make sure there's a tooltip sticky variant provided but not a tooltip sticky position
1. Make sure the 0 is not in green


components/atoms/Slider/Slider.js
```diff
@@ -50,7 +50,7 @@ const Slider = ({
 
   return (
     <div className={containerClasses}>
-      {tooltipStickyVariant && (
+      {tooltipStickyVariant && tooltipStickyPosition && (
```
CalMlynarczyk: This will coalesce to false and not render the tooltip if you pass in `0` for this prop.

RohitRajendran: Good catch!

PR #980, opened 2020-01-10
More Docs

JamesIves:
Explains the need for `eve.navigatingToLatestCard`.


components/templates/McGonagall/McGonagall.mdx
```diff
@@ -243,6 +243,41 @@ name: {
 },
 ```
 
+Actions run on the transitions between one state to another. When a user returns to a flow McGonagall will also sweep through each state for questions that have already been answered. Depending on the purpose of the action it may be required to check if `eve.navigatingToLatestCard` is false to prevent data from being incorrectly cleared, you can view an example of this below in the `clearSpendingTypeDependencies` function.
+
+```javascript
+const stateOptions = {
+  actions: {
+    clearSpendingTypeDependencies: assign((ctx, eve) => {
```
RohitRajendran: I think a better example was the mark as incomplete  one

RohitRajendran: This doesn't really need it

PR #993, opened 2020-01-14
Standardizes Z-Index Values

JamesIves:
Makes a change which standardizes z-indexes on the following scale:

```sass
// z-index scale in ascending order.
$z-index-auto: auto;
$z-index-below: -1;
$z-index-default: 1;
$z-index-popup: 4000;
$z-index-dialog: 5000;
$z-index-dropdown: 6000;
$z-index-overlay: 7000;
$z-index-menu: 8000;
$z-index-modal: 9000;
$z-index-toast: 10000;
```

I've started to these these in client and so far so good. These get exported via Webpack so you can import either the Sass variables, JavaScript variables, or you can just use the provided helper classes. I figured all three was appropriate as we do something similar for colors.

I've also created a story which stacks these items, that way we can catch regressions early if something ends up changing with these values (though they shouldn't change). I will need to update the main README file if everyone is ok with this. 

* Load up the `utilities/Styles/zindex` story to see this in action.

There will be a subsequent change that will come soon after to fully update the dotcom and client sites to use these variables. 

There's some occurrences where there's a single digit increment on the z-index. I think using Sass to do this as an exception is a fine pattern so long as it's by `1` and not `1000` or something ridiculous. 


components/utilities/Styles/zindex.story.js
```diff
@@ -0,0 +1,145 @@
...
+  <div className="uic--position-relative">
+    <div
+      style={{
+        ...sharedStyles,
+        backgroundColor: colors.midnight,
+      }}
+      className="uic--z-index-below uic--position-absolute"
+    >
+      .uic--z-index-below <br />
+      $z-index-below
```
RohitRajendran: Can we also specify what values each of them are too?

JamesIves: Yep, I'll actually go as far as to making this dynamic based on what's in the javascript object.

JamesIves: Improved this, added Sass variable names, and the value.

constants/js/zindex.js
```diff
@@ -0,0 +1,13 @@
+/** z-index values accessible in JavaScript. */
+export const zindex = {
+  'z-index-below': '-1',
+  'z-index-auto': 'auto',
+  'z-index-default': '1',
+  'z-index-popup': '4000',
+  'z-index-dialog': '5000',
+  'z-index-dropdown': '6000',
```
RohitRajendran: Dropdown is higher than a popup or dialog? Shouldn't it just be a little bit higher than default?

JamesIves: Yeah, I think this makes sense because they are primarily interactive. Drop-downs should never be covered by anything with the exception of anything that is designed to overlay content. 

The term popup/dialog here is used for things like tooltips and explanation text.

CalMlynarczyk: I'd just call dialog tooltip so it's consistent with what we refer to those as; I had to go look up what you meant by dialog. And get rid of the popup one since it's not used. It's confusing whether that's the same as a modal.

JamesIves: @CalMlynarczyk Good call - I've updated this.

@RohitRajendran I've added the docs you requested.

components/utilities/Styles/zindex.mdx
```diff
@@ -0,0 +1,94 @@
...
+    <td>
+      Primarily used for elements which overlay content. For example the Cabinet
+      component.
+    </td>
+    <td>7000</td>
+  </tr>
+  <tr>
+    <td>z-index-menu</td>
+    <td>
+      Primarily Used for navigation bars which appear at the top of the screen.
```
RohitRajendran: This still sounds like it should be lower than an overlay

JamesIves: The navigation bar isn't part of content though.

JamesIves: I changed the description to be a little more clear in this scenario.

CalMlynarczyk: I don't know which way I prefer. I think the navbar itself should probably be under the overlay since it is fixed and will probably squish the area where the modal can fit. However, the mobile nav menu itself might be in the same vein as a dropdown and should appear over top of the overlay if it is activated.

JamesIves: Personally I prefer menu to be higher than overlay as having a menu overlaid by content besides a full page modal doesn't make sense to me. The use case for this would be a mobile global nav should appear over the cabinet (overlay) if the user clicks the button. Anything higher than menu should require some form of interaction to close (modal or toast).

JamesIves:
@CalMlynarczyk @RohitRajendran Let me know if you guys have additional requests/changes.

PR #1029, opened 2020-01-23
[APP-4371] Add multi-line functionality to Input component

CalMlynarczyk:
### Description
Design requires a free-form text field for a new page. Since the field would be so similar to the existing `Input` component, I just added the capability to that component using a flag. UXD specified that the field should be resizable vertically.

### Testing Instructions
Go to the _multiline_ example under the `Input` component.

### Screenshots
Screen Shot 2020-01-23 at 11 59 51

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4371)


angelojcruzjr:
I think the logic for creating this checks out but if we are going to add this we should redo the layout logic to avoid overflow on the Input label (see below):
Screen Shot 2020-01-23 at 11 25 17 AM

CalMlynarczyk:
@angelojcruzjr Hmm, yea. Built this in FF. Didn't realize it would spaz out in Chrome. I'll look into that.

CalMlynarczyk:
@angelojcruzjr Think I resolved the issue.

components/atoms/Input/Input.scss
```diff
@@ -6,19 +6,36 @@ $input-data-value-color: rgba(255, 255, 255, 0);
...
+      border: 1px solid $pink;
+
+      &:focus,
+      &:active {
+        border: 1px solid $pink;
+      }
+    }
+
+    textarea {
+      outline: 1px solid $pink;
```
JamesIves: Does this still work correctly when keyboard tabbing with the focus-visible polyfill?

yarn.lock
```diff
@@ -4828,7 +4828,7 @@ debug@3.1.0, debug@=3.1.0:
   dependencies:
     ms "2.0.0"
 
-debug@3.2.6, debug@^3.1.0, debug@^3.2.5, debug@^3.2.6:
+debug@3.2.6, debug@^3.1.0, debug@^3.2.5:
```
JamesIves: Are these changes intentional?

components/atoms/Input/Input.mdx
```diff
@@ -41,6 +41,40 @@ import {Input} from '@unitedincome/components';
 />
 ```
 
+### Multi-line inputs
+
+You can create a multi-line input similar to an HTML text area. Multi-line inputs do not support the `mask` prop.
```
JamesIves: This is nitpicky, but can we place this below the header for `Masks`? It just reads better with prior context on what that prop does.

JamesIves:
JK You fixed the coverage right before I commented

yarn.lock
CalMlynarczyk: Yea, I updated them in `develop` as well.

components/atoms/Input/Input.scss
CalMlynarczyk: As far as I tested it did.

PR #1033, opened 2020-01-29
[APP-4371] Auto-resize multi-line input

CalMlynarczyk:
### Description
Monique and I discussed how to resolve the weird display issues with multi-line inputs without having to significantly re-architect the input component or create a brand new component. We decided to not to allow user resizing of the field. Instead, it auto resizes vertically as more rows are entered into the field.

### Testing Instructions
1. Go to the multiline input story
1. Enter multiple lines of text into the text field
1. The field should grow or shrink vertically with the number of text rows

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4371)


package.json
```diff
@@ -119,5 +119,8 @@
     "react-text-mask": "5.4.3",
     "react-tooltip": "3.11.2",
     "xstate": "4.7.6"
+  },
+  "dependencies": {
+    "react-textarea-autosize": "7.1.2"
```
RohitRajendran: Any reason this isn't under devDependencies like the rest of the packages?

RohitRajendran:
Jan-29-2020 15-22-07
I don't know if this is just because it's in storybook but scrolling can get wonky with the multiline

CalMlynarczyk:
@RohitRajendran I think that is caused by the Storybook accessibility auditor. I couldn't reproduce that if I popped out the canvas to its own tab.

package.json
CalMlynarczyk: No, moved it to match

PR #1050, opened 2020-02-06
Generate Unique Ids for React Elements

angelojcruzjr:
Changed instances where we simply use an array `index` or `value` for a react `key` prop to instead be a unique id via [shortid](https://www.npmjs.com/package/shortid).

Ensure all components behave as expected both in storybook and `client`. Still trying to verify that this would not be a breaking change.


RohitRajendran:
Won't this create new ids on every render?

angelojcruzjr:
> Won't this create new ids on every render?

When there's state no. I am still trying to make sure that's the case for `Pure` and `Functional` components, though.

angelojcruzjr:
Update. This will only be useful for stateful components. Until we use hooks. Ugh. 😭

CalMlynarczyk:
Can you give a description of the problem you are trying to solve with these changes? At a minimum, I think generating the collection in the constructor will fail in the case when the incoming props change.

PR #1057, opened 2020-02-12
Input 'validate' Prop

angelojcruzjr:
Created a new `validate` prop for the `Input` component. The idea here is that a user could pass in a simple function that returns either an empty string (valid input) or an error message and this would automatically grant them a few things, namely:

- Input validation based on the error message
- Error message automatically being set to the message of the `validate` func

All while not breaking existing functionality.

Download the branch and play with the new `Input` story labeled `validate`

For reference, this is what is passed to the `validate` story:

```javascript
validate={(input) => {
    if (input.length < 4) {
        return '';
    } else {
        return 'Please enter a valid small string';
    }
}}
```

I realize that the `maxLength` prop could accomplish the above, this was just meant to be a quick example.


components/atoms/Input/Input.js
```diff
@@ -131,20 +138,35 @@ class Input extends Component {
...
     let InputType = 'input';
     let prependCharacter = prepend;
     let appendCharacter = append;
     let inputLabel = label;
+    let errorMsg = validationErrorMsg;
+
+    // Required should take precedence
+    if (required !== undefined) {
+      requiredAttr = required;
+    } else if (validateFunctionPassed) {
```
RohitRajendran: Why do you make requiredAttr true if the validate function is passed?

components/atoms/Input/Input.js
```diff
@@ -131,20 +138,35 @@ class Input extends Component {
...
+      requiredAttr = true;
+    }
+
+    if (validateFunctionPassed) {
+      errorMsg = validate(value);
```
RohitRajendran: We seem to be validating twice. We just save the message in state when `validate` is called within `isValid`

components/atoms/Input/Input.js
```diff
@@ -402,6 +422,8 @@ Input.propTypes = {
   showRequiredError: PropTypes.bool,
   /** Optional inline styles. */
   style: PropTypes.objectOf(PropTypes.string),
+  /** Validatioin function */
```
RohitRajendran:
```suggestion
  /** Validation function */
```

components/atoms/Input/Input.js
```diff
@@ -92,11 +96,14 @@ class Input extends Component {
       this.props.mask && maskEnum[this.props.mask].isValid
         ? maskEnum[this.props.mask].isValid(this.props.value)
         : true;
+    const validateResult = validate(this.props.value);
```
RohitRajendran: We should rename this to something like `validationErrorMessage` and save it to state.

angelojcruzjr: Updated to use state in https://github.com/UnitedIncome/components/pull/1057/commits/92894235c09929140e1371c053fbe2cd00230fbf.

components/atoms/Input/Input.js
angelojcruzjr: Updated in https://github.com/UnitedIncome/components/pull/1057/commits/92894235c09929140e1371c053fbe2cd00230fbf.

components/atoms/Input/Input.js
angelojcruzjr: Updated so that we don't make this call anymore as of https://github.com/UnitedIncome/components/pull/1057/commits/92894235c09929140e1371c053fbe2cd00230fbf.

components/atoms/Input/Input.js
angelojcruzjr: In my mind if I am going to go through the effort of creating a validation function for an input box I am going to want that data aka make it required by default. I wrote this though such that if you pass `required=false` along with a `validate` function it will override this behavior and not require the input. Open to different interpretations, though.

components/atoms/Input/Input.js
```diff
@@ -83,7 +83,11 @@ class Input extends Component {
    * @return {boolean} - Returns true or false depending on the validity.
    */
   isValid() {
-    const {isValid = () => true, hideValidity = () => true} = this.props;
+    const {
+      isValid = () => true,
+      validate = () => false,
```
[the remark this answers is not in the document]
angelojcruzjr: We reference the state as of https://github.com/UnitedIncome/components/pull/1057/commits/92894235c09929140e1371c053fbe2cd00230fbf so this no longer matters.

components/atoms/Input/Input.js
RohitRajendran: I don't think that's necessarily true and something that we want to assume. A field could be optional but once there's a value in it, we want to make sure that the value is valid.

components/atoms/Input/Input.js
```diff
@@ -94,9 +95,11 @@ class Input extends Component {
         : true;
 
     return (
-      (isValid(this.props.value) && maskValidation) ||
-      (hideValidity() && isActive) ||
-      isEmpty
+      // eslint-disable-next-line eqeqeq
+      this.state.validationErrorMessage == false &&
```
RohitRajendran:
```suggestion
      !this.state.validationErrorMessage &&
```
To make it simpler

components/atoms/Input/Input.js
```diff
@@ -204,26 +225,40 @@ class Input extends Component {
...
             : e.target.value,
-        );
+        ),
+      );
+    }
+
+    if (validateFunctionPassed) {
+      onChangeFuncs.push((e) =>
+        this.setState({
+          validationErrorMessage: validate(e.target.value),
```
RohitRajendran: Why set state for error message here and have to run validate again instead of just inside the component's isValid

components/atoms/Input/Input.js
```diff
@@ -94,9 +95,11 @@ class Input extends Component {
         : true;
 
     return (
-      (isValid(this.props.value) && maskValidation) ||
-      (hideValidity() && isActive) ||
-      isEmpty
+      // eslint-disable-next-line eqeqeq
```
matthewheck: You should practically never need to disable eslint rules.

angelojcruzjr: Modified logic so we don't disable the rule in https://github.com/UnitedIncome/components/pull/1057/commits/f1ea178f7568691f3dec4e92ea56a30d545ad6ca.

components/atoms/Input/Input.js
angelojcruzjr: So I was doing it this way because setting state in the `isValid` would trigger a rerender and lead to an infinite loop of rendering. Realized that we can just make this a PureComponent to utilize the shallow render, though. Updated accordingly in https://github.com/UnitedIncome/components/pull/1057/commits/f1ea178f7568691f3dec4e92ea56a30d545ad6ca.

components/atoms/Input/Input.js
angelojcruzjr: Updated in https://github.com/UnitedIncome/components/pull/1057/commits/f1ea178f7568691f3dec4e92ea56a30d545ad6ca.

components/atoms/Input/Input.js
angelojcruzjr: Updated so that we don't default to `required={true}` in https://github.com/UnitedIncome/components/pull/1057/commits/f1ea178f7568691f3dec4e92ea56a30d545ad6ca.

components/atoms/Input/Input.js
```diff
@@ -145,6 +154,13 @@ class Input extends Component {
     let prependCharacter = prepend;
     let appendCharacter = append;
     let inputLabel = label;
+    let errorMsg = '';
```
RohitRajendran:
```suggestion
    const errorMsg = validationErrorMsg || this.state.validationErrorMessage;
```

RohitRajendran:
We should also update the input readme with an example on how to use this and a note that those old props will be getting depreciated.

components/atoms/Input/Input.js
angelojcruzjr: Changed to this in https://github.com/UnitedIncome/components/pull/1057/commits/df357a9e008f8112d5e7416c7265c646fa0b582c.

angelojcruzjr:
> We should also update the input readme with an example on how to use this and a note that those old props will be getting depreciated.

Got rid of the story and provided some how-to language in the `Input` story in https://github.com/UnitedIncome/components/pull/1057/commits/df357a9e008f8112d5e7416c7265c646fa0b582c.

PR #1065, opened 2020-02-13
InfoBox & InfoBoxFormattedRows -> TypeScript

angelojcruzjr:
TypeScript-ified the `InfoBox` and `InfoBoxFormattedRows` components.

Pull down the branch and ensure the components still work and behave as expected locally.

A few strange things I encountered:

1. In testing, I have a test that returns an array of `ReactNode`s that I try to find the `.length` of. TS did not like this unless I casted the variable initially to type `any`. This will undoubtedly come up in other parts of the component library and `client` as well so we should figure out what the best way to address this is.
2. `InfoBoxFormattedRows` has a property (`rowData`) that is an array of objects. I was able to type it correctly but I don't think the expected properties of each object shows up appropriately in the Storybook docs. Hopefully there is a workaround for this but I wasn't able to find one in my brief ~15 mins of searching.


components/atoms/InfoBox/InfoBox.spec.tsx
```diff
@@ -22,11 +22,13 @@ test('InfoBox - renders correctly', (t) => {
...
-  t.equals(
-    component.find('.uic--info-box-information-area').prop('children').length,
-    2,
-    'Should render a content and footer child',
-  );
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const jsxPropsArray: any = component
+    .find('.uic--info-box-information-area')
+    .prop('children');
```
angelojcruzjr: Weird thing #1

components/atoms/InfoBoxFormattedRows/InfoBoxFormattedRows.tsx
```diff
@@ -1,6 +1,23 @@
...
+import React, {FC} from 'react';
+
+type rowDataProps = {
+  /** Optional unique key for a row of data. */
+  key?: string;
+  /** Label that will be displayed on the left side of the row. */
+  label: string;
+  /** Value that will be displayed on the right side of the row. */
+  value: string;
+};
```
angelojcruzjr: Weird thing 2) - these props don't show up in the Storybook docs tab

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -1,18 +1,40 @@
...
 /** @module InfoBox */
-import PropTypes from 'prop-types';
 import React, {Component} from 'react';
 import './InfoBox.scss';
 import classNames from 'classnames';
 import {formatEllipsis} from '../../utilities/FormatUtils/FormatUtils';
 
+type InfoBoxProps = {
+  /** The main title that will displayed near the top of our InfoBox. */
+  title?: string;
```
CalMlynarczyk: In order to match the previous prop types definitions, this shouldn't be optional.

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -28,7 +50,7 @@ class InfoBox extends Component {
    * @param {object} event - The event object.
    * @returns {function} Function to be ran
```
CalMlynarczyk: We don't need types in TypeScript JSDoc sections.

components/atoms/InfoBoxFormattedRows/InfoBoxFormattedRows.tsx
CalMlynarczyk: I see the reference to them in the doc tab, but the entire type definition isn't going to show up there because it's not the type of the component props.

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -115,11 +132,17 @@ class InfoBox extends Component {
           )}
           <div
             className="uic--info-box-title-area"
-            style={imgStyle}
+            style={
+              imageURL
+                ? {
+                    backgroundImage: `url(${imageURL})`,
+                  }
+                : {}
```
CalMlynarczyk: If we're going to inline this here, then `imgStyle` above is not longer used and should get removed.

components/atoms/InfoBox/InfoBox.spec.tsx
```diff
@@ -70,8 +72,8 @@ test('InfoBox - renders correctly', (t) => {
 
 test('InfoBox - state and hover updated correctly', (t) => {
   const testProps = {
-    title: 'This is a very, very long title',
-    content: 'Here is some content!',
+    title: 'This is a very, very long title' as string,
```
CalMlynarczyk: Implicit typing should work fine here. You don't need to cast this value to `string`.

components/atoms/InfoBox/InfoBox.spec.tsx
```diff
@@ -70,8 +72,8 @@ test('InfoBox - renders correctly', (t) => {
...
+    content: 'Here is some content!' as React.ReactNode,
```
CalMlynarczyk: Again, you do not need to cast this. A `string` is a valid `ReactNode`.

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -41,10 +63,13 @@ class InfoBox extends Component {
    * @param {Event} event - Event obj
    * @returns {undefined}
    */
-  handleMouseOver(event) {
+  handleMouseOver(event: React.MouseEvent): void {
     event.stopPropagation();
 
-    if (this.props.title.length > this.props.titleCharLimit) {
+    if (
+      this.props.title &&
```
CalMlynarczyk: Once you change this to be required, you won't have to check its existence.

components/atoms/InfoBox/InfoBox.spec.tsx
CalMlynarczyk: A couple of things here. First, I think the original test code actually wasn't good, maybe even incorrect. I would not expect there to be two children because we don't define footer. At a minimum, we should have made sure the second child was `undefined` instead of only checking the array length.

Second, I think the better way to do this is to call the `.children()` function on the found node instead of `.prop('children')`. Enzyme takes care of wrapping the children nodes so we can analyze them easier in test code, including handling the typing better for us in the case that nodes aren't found at runtime. You'll notice once you change this code that the children length goes down to `1` (meaning we have to adjust the test case as well) because it accounts for one child node being `undefined`. I think that's both the correct test expectation and it simplifies the test code.

components/atoms/InfoBoxFormattedRows/InfoBoxFormattedRows.tsx
```diff
@@ -9,7 +26,9 @@ import React from 'react';
  * @param {Object} props Object containing the component props
  * @returns {JSX} InfoBoxFormattedRows JSX
  */
-const InfoBoxFormattedRows = (props) => {
+const InfoBoxFormattedRows: FC<InfoBoxFormattedRowsProps> = (
+  props: InfoBoxFormattedRowsProps,
```
CalMlynarczyk: If you want, you can use destructured args here like we do with most of our functional components. The TS compiler will know that the props match the props type you provided.

angelojcruzjr: Desctructured in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: Changed `title` to required in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f.

components/atoms/InfoBox/InfoBox.spec.tsx
angelojcruzjr: Made implicit in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f.

components/atoms/InfoBox/InfoBox.spec.tsx
angelojcruzjr: Made implicit in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: Yup this was leftover on accident. Deleted in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: Got rid of references to types in JSDocs in https://github.com/UnitedIncome/components/pull/1065/commits/fb5e3e9f6d5cbe985fff368d61e24c94a10ede59.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: Made required in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f.

components/atoms/InfoBoxFormattedRows/InfoBoxFormattedRows.tsx
angelojcruzjr: So instead of making `rowData` a type and plugging in an array of them, I explicitly typed what the object should look like in the broader `props` object that gets passed in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f. I now see those types displayed in the docs which is the current behavior so I will follow this style from now on.

components/atoms/InfoBox/InfoBox.spec.tsx
angelojcruzjr: You are correct the previous test was reading a null value. The right behavior I think is to check that the text we passed gets rendered so I made that the check in https://github.com/UnitedIncome/components/pull/1065/commits/2ffba27101bfddc39c641c7e0a78942746a51d3f.

CalMlynarczyk:
@angelojcruzjr Just noticed this was still in draft. Do you still want me to review as if it's ready to go?

angelojcruzjr:
> @angelojcruzjr Just noticed this was still in draft. Do you still want me to review as if it's ready to go?

Yeah sure thing! Might as well get things together on your branch.

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -1,18 +1,40 @@
...
+};
+
 /**
  * Renders a basic card with a title component on top and
  * description/more information below the title portion.
  */
-class InfoBox extends Component {
+class InfoBox extends Component<InfoBoxProps, {showFullTitle: boolean}> {
+  /** @inheritdoc */
+  public static defaultProps = {
```
JamesIves: Do we want to start defining defaultProps like this across all of our components? I'd much prefer to keep this consistent.

Depending on the outcome of this discussion we should probably update the docgen example and enforce that all future additions/updates adhere to this standard.

angelojcruzjr: Good question. I'm pretty impartial to whatever implementation we use but I agree it should be uniform. @CalMlynarczyk any thoughts?

CalMlynarczyk: I ultimately want to use the class properties syntax you added here. I know we don't do it anywhere else yet but I think it helps keep component code organized within its class definition.

However, when we change this to a hooks-based implementation (looking likely that is what will happen), we will probably have to go back to assigning `defaultProps` in a separate statement again. I don't know of another way to do that for functional components.

PR #1068, opened 2020-02-13
Atoms/Illustrations -> TypeScript -> Hooks

JamesIves:
This is a combination of two projects that demonstrates both TypeScript and React Hooks. In this branch I've migrated all `atoms/illustration` components to be typed, I've also updated the component (and tests) to use React Hooks and the react-testing-library package. I've also used `aria-labelledby` to make these svgs more accessible. 

* Related testing principals: https://testing-library.com/docs/dom-testing-library/api-queries#bytestid

* Everything should behave as normal, you should be able to click the illustration and see it toggle correctly.

We're going to be doing a similar presentation on hooks soon, so this is just prep work for that.


components/atoms/illustrations/CarCircleIllustration/CarCircleIllustration.tsx
```diff
@@ -0,0 +1,516 @@
...
+  illuminate,
+  color,
+  draft,
+  width,
+  height,
+  className,
+  style,
+  id,
+}: CarCircleIllustrationProps) => {
+  const [illuminated, setIllumination] = useState(false);
```
dylanbyars: Ultra nitpick: change `setIllumination` to `setIlluminated` since that setter is setting the `illuminated` variable.

components/atoms/illustrations/CarCircleIllustration/CarCircleIllustration.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+test('CarCircleIllustration - renders', (t) => {
+  const component = render(<CarCircleIllustration />);
+
+  t.equals(
+    component.container.querySelectorAll('svg').length,
+    1,
+    'Should load the CarCircleIllustration component.',
+  );
+
+  t.true(getByTestId(component.container, 'purple'), 'Purple car illustration');
```
labichn-cof: Can this be `component.getByTestId('purple')` instead?

JamesIves:
Also implemented react-testing-library. The goal is to eventually deprecate the use of Enzyme.

components/atoms/illustrations/CarCircleIllustration/CarCircleIllustration.spec.tsx
JamesIves: Done

components/atoms/illustrations/AccountCircleIllustration/AccountCircleIllustration.tsx
```diff
@@ -0,0 +1,228 @@
...
+};
+
+/** Renders an account illustration. */
+const AccountCircleIllustration: FC<AccountCircleIllustrationProps> = ({
+  fill,
+  width,
+  height,
+  className,
+  style,
+}: AccountCircleIllustrationProps) => (
```
CalMlynarczyk: I don't believe you need to annotate the props type here when you already included it in the generic `FC<>` declaration above.

components/atoms/illustrations/AccountCircleIllustration/AccountCircleIllustration.spec.tsx
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    t.equals(
+      component.container.querySelectorAll('svg').length,
+      1,
+      'Should load the AccountIllustration component.',
+    );
+
+    t.true(component.getByTestId('default'), 'Default illustration');
+
+    t.end();
```
CalMlynarczyk: `t.end()` needs to go in the `finally` block. If the `catch` block is executed, the entire test suite will fail because this test will never get exited. I also don't know if the `try`/`catch` usage here is necessary. I only started doing that with `async` code because unexpected promise rejections behave differently from thrown exceptions and will break the final test results when they occur.

JamesIves: Ah good point - I can make that adjustment.

components/atoms/illustrations/AccountCircleIllustration/AccountCircleIllustration.tsx
JamesIves: I'll give that a shot, if TypeScript doesn't get upset I'll make the changes across the board.

JamesIves:
We'd still need to actually test the implementation of the onClick handler which ultimately wouldn't reduce the amount of tests. With this new approach to testing we're not calling the click handler directly, instead we're simulating a click on the component and testing the visible result to the user. 

We should also have a discussion during Friday's meeting about how much we want to share between components, and when it makes sense to branch something out to make it sharable. I'd also like to come to a consensus about where these should be stored in the file structure. 

**For example:** 

`~hooks/state/setIllumination.state.ts`
`~hooks/effects/setValue.effect.ts`

components/atoms/illustrations/AccountCircleIllustration/AccountCircleIllustration.spec.tsx
labichn-cof: I'd rather the cleanup just show up in the finally block, since all tests should have try/catches anyway. It would be a pretty thin wrapper, though, so I could go either way.

labichn-cof:
> Also, we should consider creating an HOC or testing the `setIllumination` hook separately so that we can wrap all of the illustrations that have that functionality. That way we don't need to re-implement and test it on _every_ component. We don't have to do that now, but it made ultimately save you some effort here if you do.

Just a few clarifications--`setIllumination` isn't a hook, it's a function that updates the illumination state value. `useState` is the hook, and does not need to be tested just like `this.setState` doesn't need to be tested.

Also, HOCs should really be used only when necessary due to the way they extend the depth of the render tree--hooks are one of the solutions to that problem: https://reactjs.org/docs/hooks-faq.html#do-hooks-replace-render-props-and-higher-order-components.

components/atoms/illustrations/AccountCircleIllustration/AccountCircleIllustration.tsx
JamesIves: Turns out you _do_ need it otherwise `react/prop-types` errors the linter.

CalMlynarczyk: I think we should adjust that linting rule for these cases then, _not_ the code itself.

PR #1070, opened 2020-02-19
TypeScript Module Aliasing

JamesIves:
Re-implements module aliasing into TypeScript. Thanks to @labichn-cof we were able to get `yarn test` working.

* Run `yarn test` and `yarn testSingle`, both will pass tests. Run the tests on the `Box.tsx` file as I've added path alaises to it too.
* VSCode should properly resolve the aliasing for you, magic! 

This will allow us to keep all of our docs intact.


package.json
```diff
@@ -118,6 +118,7 @@
     "tape": "4.13.0",
     "text-mask-addons": "3.8.0",
     "ts-node": "8.6.2",
+    "tsconfig-paths": "^3.9.0",
```
matthewheck: drop the ^

package.json
```diff
@@ -118,7 +118,7 @@
     "tape": "4.13.0",
     "text-mask-addons": "3.8.0",
     "ts-node": "8.6.2",
-    "tsconfig-paths": "^3.9.0",
+    "tsconfig-paths": "3.9.0",
```
RohitRajendran: Need to run yarn and commit lock file changes

PR #1071, opened 2020-02-20
Atoms/Loader -> TypeScript

JamesIves:
Converts the `Loader` component to use TypeScript and react-testing-library.

* Should be the same. Code changes only.


components/atoms/Loader/Loader.spec.tsx
```diff
@@ -0,0 +1,22 @@
...
+  try {
+    const component = render(<Loader />);
+
+    t.equals(
+      component.container.querySelectorAll('svg').length,
+      2,
+      'Should load the Loader component SVG tags.',
+    );
+
+    t.end();
```
CalMlynarczyk: `t.end()` needs to go in the `finally` block. I also don't think the `try`/`catch` stuff is necessary for synchronous tests, as `tape` will treat uncaught exceptions as failed tests automatically.

labichn-cof: Doesn't a single, uncaught failing test cause cascading failures in other tests? It can be hard to track down what actually went wrong when that happens.

CalMlynarczyk: An uncaught exception "crashes" the test process. You still get results up until that test though. There is a [tape-catch package](https://github.com/michaelrhodes/tape-catch) that we could use to run the remaining tests even if any error out.

JamesIves: It works well in the `finally` block.

PR #1072, opened 2020-02-20
Atoms/icons/arrow-ellipsis-icon -> TypeScript

ortizlu:
convert one file to TS just to make sure I'm doing it right

check to make sure conversion was done right.


components/atoms/icons/ArrowEllipsisIcon/ArrowEllipsisIcon.spec.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+    t.equals(
+      component.container.querySelectorAll('svg').length,
+      1,
+      'Should load the ArrowEllipsisIcon component.',
+    );
+
+    t.true(component.getByTestId('default'), 'Default Icon');
+
+    t.end();
```
CalMlynarczyk: See https://github.com/UnitedIncome/components/pull/1071/files#r382311203

components/atoms/icons/ArrowEllipsisIcon/ArrowEllipsisIcon.tsx
```diff
@@ -1,12 +1,27 @@
...
+};
+
+/** Renders an account illustration. */
+const ArrowEllipsisIcon: FC<ArrowEllipsisIconProps> = ({
+  width,
+  height,
+  fill,
+  className,
+  style,
+}: ArrowEllipsisIconProps) => (
```
CalMlynarczyk: You shouldn't need to annotate the props type here since you already included it in the generic `FC<>` declaration above.

package.json
```diff
@@ -138,5 +138,8 @@
     "react-text-mask": "5.4.3",
     "react-tooltip": "3.11.2",
     "xstate": "4.7.6"
+  },
+  "dependencies": {
+    "@testing-library/react": "^9.4.0"
```
CalMlynarczyk: This should go in `devDependencies` and use an exact version.

JamesIves: I committed this to the `feature/typescript` branch so you can also just update from that and it will already be included.

components/atoms/icons/ArrowEllipsisIcon/ArrowEllipsisIcon.tsx
```diff
@@ -23,6 +38,7 @@ const ArrowEllipsisIcon = ({width, height, fill, className, style}) => (
     }
     viewBox="0 0 24 24"
     transform="rotate(0)"
+    data-testid="default"
```
JamesIves: I'm not really sure we need to use `testid` here. This is only useful for illustrations which toggle between different svgs. It's probably redundant.

components/atoms/icons/ArrowEllipsisIcon/ArrowEllipsisIcon.tsx
JamesIves: I had issues with `yarn lint` throwing an error on prop-types if it's not included. I'm not sure if that's something we can turn off or not.

PR #1097, opened 2020-02-27
Atoms/Icons -> typescript

ortizlu:
Converted all of the icons to typescript, updated all tests to use testing-library-react, and updated all of the pathnames. 

make sure conversion is 1 to 1 for all components.

N/A


components/atoms/icons/ArrowEllipsisIcon/ArrowEllipsisIcon.tsx
```diff
@@ -14,7 +14,10 @@ type ArrowEllipsisIconProps = {
   style?: React.CSSProperties;
 };
 
-/** Renders an arrow ellipsis icon. */
+/** Displays the ArrowEllipsisIcon component.
+ * @param props - Component props.
```
JamesIves: I'm not sure these comments are really necessary here as we annotate the prop types.

RohitRajendran: Yeah agreed!

CalMlynarczyk: Same

ortizlu: Do we still want the `@returns` comments or should we just keep the description of the function only

JamesIves: I think just the description

PR #1099, opened 2020-02-28
Bubble Selector -> TypeScript

ortizlu:
change Bubble Selector Component to ts.
Added Pure export of icons to allow props to be used in storybook documentation.

Check to make sure conversion is 1-1


components/atoms/BubbleSelector/BubbleSelector.tsx
```diff
@@ -1,9 +1,31 @@
...
+
+type OptionsType = Array<OptionsConfig>;
+
+type BubbleSelectorProps = {
+  /** Collection of option labels and optional colors */
+  options: OptionsType;
+  /** Currently selected value */
+  value: string;
+  /** Handler for when the selected value changes */
+  onChange: (value: string | undefined) => void;
```
RohitRajendran: Could value ever actually be undefined?

components/atoms/icons/InfoIcon/InfoIcon.tsx
```diff
@@ -3,13 +3,13 @@ import React, {FC, memo} from 'react';
 import {colors} from '~constants/js/colors';
 
 type InfoIconProps = {
-  /** The width of the arrow with unit sizing (px, rem, etc). */
+  /** The width of the infoIcon with unit sizing (px, rem, etc). */
```
RohitRajendran: Good catch!

components/atoms/BubbleSelector/BubbleSelector.tsx
ortizlu: lol good point

PR #1103, opened 2020-03-04
[APP 4720] - Life Insurance selection screen

ortizlu:
Added two new illustrations for the life insurance selection screen - UmbrellaCircleIllustration and InsuranceCircleIllustration

check to make sure they look nice and pretty.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4720)


components/atoms/illustrations/InsuranceCircleIllustration/InsuranceCircleIllustration.mdx
```diff
@@ -0,0 +1,68 @@
+import {Preview, Props} from '@storybook/addon-docs/blocks';
+import {PureInsuranceCircleIllustration as InsuranceCircleIllustration} from './InsuranceCircleIllustration.tsx';
+
+# InsuranceCircleIllustration
+
+The InsuranceCircleIllustration component creates an insurance illustration using inline SVG. This illustration is commonly used within McGonnagall introduction cards.
```
JamesIves: What is an `InsuranceIllustration`? Can we be more descriptive as to what this actually is? Same goes for the a11y description/title. If a screen reader were to hear `Insurance icon with sword and shield` it wouldn't really mean much to them.

ortizlu: what do you think would be a better fitting name? Document Illustration? Paper Illustration?

JamesIves: DocumentIllustration would work!

PR #1142, opened 2020-03-18
[APP-4505] InfoBox Component Updates

angelojcruzjr:
Made slight styling tweaks to the `InfoBox` component based on feedback from Monique. Additionally, added a new `isInactive` mode to the component that allows the component to be displayed as "inactive" aka a lower opacity.

Download the codebase and run locally. I added a new story for the `isInactive` state - ensure that the card color changes appropriately - the cards should look similar to the current grayed-out cards in the draft plan section.

Screen Shot 2020-03-19 at 10 47 39 AM

This work is being pushed as part of the JIRA ticket below (rather than being it's own story). It is leftover work that we never got to when the component was first created.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/APP-4505)


components/atoms/InfoBox/InfoBox.scss
```diff
@@ -3,17 +3,15 @@ $background-size: 17rem;
 $container-min-height: 2.4rem;
 
 .uic--info-box {
-  border: 1px solid $light-gray;
+  border: thin solid $light-gray;
```
JamesIves: I'm not a fan of using `thin` here. Can we not just use a decimal number as the pixel value?

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -15,6 +16,8 @@ type InfoBoxProps = {
   imageURL?: string;
   /** Optional footer JSX to be displayed below the content. */
   footer?: React.ReactNode;
+  /** Optional inactive marker, if true sets the background of the card to grey */
+  inactive?: boolean;
```
JamesIves: We should call this `isInactive`

components/atoms/InfoBox/InfoBox.spec.tsx
```diff
@@ -112,3 +119,36 @@ test('InfoBox - state and hover updated correctly', (t) => {
 
   t.end();
 });
+
+test('InfoBox - inactive color', (t) => {
```
JamesIves: Can we migrate these tests to RTL? The lift wouldn't be particularly large, and as we're adding a bunch of tests anyway this would be a great time to migrate this component.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: Updated to `isInactive` in 889d8c4.

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: So I originally used thin because when I was testing these changes in the actual app I found that the border did not always render for some reason. I found this odd and looked it up and apparently it is a known issue with chrome (in using 1px for borders, specifically). I played around with values and 1.3px was the smallest that seems to always consistently render the borders correctly. Updated in 889d8c4.

components/atoms/InfoBox/InfoBox.spec.tsx
angelojcruzjr: Sure thing - will start working on porting over the tests now.

components/atoms/InfoBox/InfoBox.scss
CalMlynarczyk: Were you seeing that when zoomed out in your browser? I've never seen pixel rounding issues on whole pixel values in any other case.

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -102,11 +106,17 @@ class InfoBox extends Component<InfoBoxProps, {showFullTitle: boolean}> {
       },
       className,
     );
+    const backgroundColor = isInactive ? colors.slate : colors.midnight;
+
+    const cardStyling = {
+      backgroundImage: imageURL ? `url(${imageURL})` : 'none',
+      backgroundColor,
```
CalMlynarczyk: Do we need this rule on both the inner and outer card elements?

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -3,17 +3,15 @@ $background-size: 17rem;
 $container-min-height: 2.4rem;
 
 .uic--info-box {
-  border: 1px solid $light-gray;
+  border: 1.3px solid $light-gray;
```
CalMlynarczyk: This is displaying weird in Chrome and FF. We should just use a `1px` whole value for this. I can't recreate the disappearing border issue you mentioned before.

components/atoms/InfoBox/InfoBox.story.jsx
```diff
@@ -23,11 +23,19 @@ store.subscribe(() => {
...
+  imageURL,
+  content,
+  isInactive,
+  footer,
+}) => ({
   name: formName,
   title: text('title', title),
   imageURL: text('imageURL', imageURL),
   content: object('content', content),
+  isInactive: object('isInactive', isInactive),
```
CalMlynarczyk: This would be easier to use as a checkbox.

components/atoms/InfoBox/InfoBox.tsx
```diff
@@ -102,11 +106,17 @@ class InfoBox extends Component<InfoBoxProps, {showFullTitle: boolean}> {
       },
       className,
     );
+    const backgroundColor = isInactive ? colors.slate : colors.midnight;
```
CalMlynarczyk: I don't think this is what we want for faded out cards. See the comp [here](https://zpl.io/bAQMOOo). You could probably accomplish this with an opaque overlay div. CSS filters unfortunately aren't well enough supported to use those.

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: Changed to 1px in https://github.com/UnitedIncome/components/pull/1142/commits/ca084cef0f7accfb7512ad5a9e691a0c5bfbf9f8. I was zoomed out like you had mentioned above - turns out it is not an issue when you have regular zoom!

components/atoms/InfoBox/InfoBox.scss
angelojcruzjr: Updated to 1px in https://github.com/UnitedIncome/components/pull/1142/commits/ca084cef0f7accfb7512ad5a9e691a0c5bfbf9f8.

components/atoms/InfoBox/InfoBox.spec.tsx
angelojcruzjr: Using RTL as of https://github.com/UnitedIncome/components/pull/1142/commits/ca084cef0f7accfb7512ad5a9e691a0c5bfbf9f8.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: Working on using opacity rather than different colors.

components/atoms/InfoBox/InfoBox.story.jsx
angelojcruzjr: Updated in https://github.com/UnitedIncome/components/pull/1142/commits/ca084cef0f7accfb7512ad5a9e691a0c5bfbf9f8.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: We now use opacity in https://github.com/UnitedIncome/components/pull/1142/commits/55a2be40cac39da71accb664424ea39fda49a8a9.

components/atoms/InfoBox/InfoBox.tsx
angelojcruzjr: It should only exist in the title area now (idea is to paint the image if we have it).

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -7,7 +7,6 @@ $container-min-height: 2.4rem;
 
   hr {
     background-color: $light-gray;
-    border: 0;
     color: $light-gray;
     height: 1px;
```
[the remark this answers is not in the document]
angelojcruzjr: Updated `<hr>` styling in https://github.com/UnitedIncome/components/pull/1142/commits/abe971c966407ce4f40ba2efde34d29d8b374f72 to follow browser stylesheet standards.

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -6,9 +6,10 @@ $container-min-height: 2.4rem;
   border: 1px solid $light-gray;
 
   hr {
-    background-color: $light-gray;
-    color: $light-gray;
-    height: 1px;
+    border: 0;
+    border-top: 1px solid $light-gray;
+    margin-block-end: 10px;
+    margin-block-start: 10px;
```
CalMlynarczyk: We should still use `rem`s like we do for all other spacing.

components/atoms/InfoBox/InfoBox.scss
```diff
@@ -6,10 +6,10 @@ $container-min-height: 2.4rem;
   border: 1px solid $light-gray;
 
   hr {
-    background-color: $light-gray;
     border: 0;
-    color: $light-gray;
-    height: 1px;
+    border-top: 1px solid $light-gray;
+    margin-block-end: 1rem;
+    margin-block-start: 1rem;
```
JamesIves: Don't use `margin-block-end` or `margin-block-start`. It has no support in IE: 

https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-end
https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-start

JamesIves: Another link; https://caniuse.com/#search=margin-block-end

PR #1148, opened 2020-03-20
Convert card components to TypeScript

CalMlynarczyk:
### Description
These components were not outputting correct type definitions in the published package because of an update to the TS version. Compilation worked correctly, but consuming them did not.

We also unfortunately have to move away from the resolved aliases for module imports because they do not compile correctly to their derived relative paths.

### Comments
I did not convert the test files for the sake of time. JS tests still work correctly with TS modules, although they do miss out on a little bit of type checking to make sure our tests make sense.


JamesIves:
> We also unfortunately have to move away from the resolved aliases for module imports because they do not compile correctly to their derived relative paths.

What change causes this?

CalMlynarczyk:
@JamesIves I think it was the cross-module reference types that I used.

For example:

```tsx
import QuestionCard from '~components/organisms-simple/QuestionCard/QuestionCard';

...
type CheckboxCardProps = typeof QuestionCard.propTypes & { ... };
```

This does not work because the compiled type definition file looks like this:

```tsx
import QuestionCard from '~components/organisms-simple/QuestionCard/QuestionCard';

...
declare type CheckboxCardProps = typeof QuestionCard.propTypes & { ... };
```

When you try to import that type definition file in _client_, it fails because _client_ doesn't know how to process that import path.

PR #1177, opened 2020-04-06
[APP-4809] Add de-emphasized state and remove footer on InfoBox

CalMlynarczyk:
### Description
Add a new style for a "de-emphasized" state for `InfoBox`. Also, the footer prop doesn't really make sense anymore because the paradigm kind of falls apart with the new changes. There isn't really anything to distinguish the footer from the main card content because there is not a consistent layout for the bottom part of the card. We should not revisit this until we have at least three use-cases for this part so we can make the correct abstraction.

### Testing Instructions
1. Toggle the `isInactive` and `isDeemphasized` states around in storybook on the `InfoBox` component

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4809)


CalMlynarczyk:
> Looks good! Guessing the one use of infobox with footer right now in client is what you're working to change so that will get updated along with this?

Yea, I have already moved those pieces into the `content` prop.

PR #1179, opened 2020-04-06
Add workaround for hooks error with yarn link to readme

maggie9595:
Seems like this is only happening to certain people so throwing that link in the readme. Thanks @ortizlu!

The error for future reference:
image


CONTRIBUTING.md
```diff
@@ -11,6 +11,8 @@ You can mimic publishing this repository locally by running `yarn link` in the d
 
 At this point, whenever you make changes to the component library and run `yarn build`, the code running in the other repository will automatically change.
 
+If you are getting errors with hooks after using `yarn link` (Error: Invalid hook call. Hooks can only be called inside of the boyd of a function component.), try this workaround: https://github.com/facebook/react/issues/14257#issuecomment-595183610/.
```
RohitRajendran, line 14: Can you also add what the workaround is here along with the link

PR #1202, opened 2020-04-20
[APP-4909] InterstitialScreen Component

JamesIves:
Implements the `InterstitialScreen` component.

* Take a look at it and read through the docs. It's modular enough to be able to be used in multiple different places.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-4909)


components/molecules/InterstitialScreen/InterstitialScreen.mdx
```diff
@@ -0,0 +1,244 @@
...
+  Icon={CarCircleIllustration}
+/>
+```
+
+### Steps
+
+Depending on the use case of the component you can define a series of steps that guide the user through a process. You can highlight the current step using the `current` prop.
+
+<Preview>
+  <InterstitialScreen
```
RohitRajendran, line 75: Should steps really be part of this component? It seems like something that's too specific for this component and should just be passed in.

components/molecules/InterstitialScreen/InterstitialScreen.scss
```diff
@@ -0,0 +1,48 @@
...
+  background-color: $royal;
+
+  &__inner {
+    margin: 2rem auto 3rem;
+    max-width: 76rem;
+    min-height: 100vh;
+
+    h1,
+    p {
+      -webkit-font-smoothing: antialiased;
```
RohitRajendran, line 11: Why is this needed here? We add this at the body level

components/molecules/InterstitialScreen/InterstitialScreen.scss
```diff
@@ -0,0 +1,48 @@
...
+      color: $white;
+      margin: 0.67em 0;
```
RohitRajendran, line 13: em instead of rem?

components/molecules/InterstitialScreen/InterstitialScreen.mdx
JamesIves: I think so - there's a lot of reusability to this functionality and it enforces style rules

RohitRajendran:
We also have the following interstitial screen uses that I don't think are accounted for in this component. Take a look at files that use `enter-narnia` class:
Screen Shot 2020-04-21 at 10 44 02 AM
Screen Shot 2020-04-21 at 10 42 17 AM

JamesIves:
@RohitRajendran You can now pass in children to customize screen view, this should make this usable for any custom implementations outside of the CTA interstitials;

components/atoms/icons/CheckIcon/CheckIcon.tsx
```diff
@@ -12,50 +12,99 @@ type CheckIconProps = {
...
+      </svg>
+    );
+  }
+};
 
 CheckIcon.defaultProps = {
   width: '1.3rem',
   height: '1.3rem',
   fill: '#008422',
+  hasBackground: true,
```
RohitRajendran, line 107: It seems weird to have a prop that defaults to true. Why not just make it `removeBackground`

components/molecules/InterstitialScreen/InterstitialScreen.scss
```diff
@@ -0,0 +1,52 @@
...
+    min-height: 100vh;
+
+    h1,
+    p {
+      color: $white;
+      margin: 1.67rem 0;
+    }
+
+    svg {
+      margin-left: 0.5rem;
```
RohitRajendran, line 16: This seems like it could cause issues to apply to all svgs

components/molecules/InterstitialScreen/InterstitialScreen.mdx
RohitRajendran: I see the InterstitialScreen component like the CardShell or QuestionCard component and if we wanted to make the InterstitialScreen with Step box a standard thing, it should be a separate component that builds on top of InterstitialScreen similar to how the RadioButtonCard builds on top of the QuestionCard. I can see us having several common variants of the interstitial screen and so I think this base one should be as simple as possible.

JamesIves: I'm fine with splitting them into two components

JamesIves: There is now a `StepBox` and an `InterstitialScreen` component.

components/atoms/icons/CheckIcon/CheckIcon.mdx
```diff
@@ -39,6 +39,18 @@ You can adjust the color of the icon using the `fill` prop. This prop accepts co
...
+### Removing Background
+
+If you'd like to render the check without the background you can toggle `removeBackground` to `true.`
+
+<Preview>
+  <CheckIcon height="3rem" width="3rem" removeBackground={true} />
+</Preview>
+
+```javascript
+<CheckIcon height="3rem" width="3rem" removeBackground={true} />
```
RohitRajendran, line 51: We have an eslint check that prevents from actually doing `removeBackgorund={true}` in favor of just adding `removeBackground`.

```suggestion
<CheckIcon height="3rem" width="3rem" removeBackground />
```

components/molecules/StepBox/StepBox.tsx
```diff
@@ -0,0 +1,89 @@
...
+            {currentStep.buttonLabel && currentStep.buttonLabel}
+          </Button>
+        </div>
+      )}
+    </div>
+  );
+};
+
+StepBox.defaultProps = {
+  className: 'uic--default-placeholder',
```
RohitRajendran, line 86: What is this default class for?

JamesIves: Leftover from running generate

PR #1213, opened 2020-04-24
InterstitialStepScreen and InterstitialScreen Tweaks / Action Bar

JamesIves:
* Implements the action bar to the interstitial screen. 
* Fixes the alignment so the interstitials properly center align to the screen. This is done with a combination of vh and absolute positioning as flex wasn't appropriate.
* Lots of visual tweaks to the screens, worked with Monique.
* Added the `InterstitialStepScreen` component which handles the step variant.

* Look at the screens, verify their props work correctly.

Screen Shot 2020-04-24 at 1 49 38 PM
Screen Shot 2020-04-24 at 1 49 42 PM
Screen Shot 2020-04-24 at 1 49 46 PM


components/molecules/InterstitialScreen/InterstitialScreen.tsx
```diff
@@ -14,27 +19,68 @@ type InterstitialScreenProps = {
...
+  );
+
   return (
     <div className={containerClasses} style={style}>
-      <div className="uic--interstitial-screen__inner uic--d-flex uic--flex-column uic--justify-content-center uic--align-items-center">
+      {actionBarOptions && (
+        <div className="uic--interstitial-screen__action-bar uic--w-100 uic--text-right">
+          <div className="uic--d-flex uic--align-items-center uic--justify-content-end">
+            <Button variant="link" dark {...actionBarOptions.buttonOptions}>
+              {actionBarOptions.buttonLabel}
```
RohitRajendran, line 70: The icon should be within the button. Should we just make buttonLabel JSX and require the icon to be part of it?

JamesIves: Are we sure about that? I thought the admin icon was just to indicate that the adjacent action was only performable by an admin.

RohitRajendran: Yeah it's within the button for this situation too: https://app.zeplin.io/project/589b5c319e603a741d558f10/screen/5e78ed8ff6d3346be7a09751

JamesIves: It's now clickable as part of the button.

components/molecules/InterstitialScreen/InterstitialScreen.tsx
```diff
@@ -14,27 +19,74 @@ type InterstitialScreenProps = {
...
   description,
+  actionBarOptions,
   Icon,
   style,
   title,
 }) => {
   const containerClasses = classNames(
     {
       'uic--interstitial-screen': true,
+      'uic--vh-100': true,
```
RohitRajendran, line 47: If this is shown under the nav which it seems to be for some situations like the Financial Plan Interstitial, wouldn't this cause a scrollbar unnecessarily?

And wouldn't we want min vh to support mobile better?

components/molecules/InterstitialScreen/InterstitialScreen.scss
```diff
@@ -1,14 +1,59 @@
 .uic--interstitial-screen {
   background-color: $royal;
 
+  &__action-bar {
+    padding: 1.3rem 1.5rem;
```
RohitRajendran, line 5:
```suggestion
    padding: 1.3rem 2rem;
```
Pretty sure it should be 2rem away from the side. It should line up with the edge of like where the log out button or like the analyze plan buttons are

components/molecules/InterstitialScreen/InterstitialScreen.tsx
JamesIves: Discussed over Slack

components/molecules/InterstitialScreen/InterstitialScreen.scss
```diff
@@ -1,14 +1,59 @@
...
+    &--icon {
+      margin-left: 0.8rem;
+    }
+
+    &::after {
+      background-color: $white;
+      content: ' ';
+      display: block;
+      height: 1px;
+      margin: 1.3rem 0 4.4rem;
```
RohitRajendran, line 26: It feels weird to me to be placing the bottom margin here instead of on `__action-bar`

components/molecules/InterstitialScreen/InterstitialScreen.scss
```diff
@@ -1,14 +1,61 @@
...
+      opacity: 0.3;
+      width: 100%;
+    }
+  }
+
+  &__wrapper {
+    flex: 1;
```
RohitRajendran, line 35: You should be able to use bootstrap utils for this too: `uic--flex-grow-1`

components/molecules/InterstitialStepScreen/InterstitialStepScreen.scss
```diff
@@ -0,0 +1,5 @@
+.uic--interstitial-step-screen {
+  .uic--cabinet-container {
```
RohitRajendran, line 2: Do we need to apply this just to that one specific cabinet in case another cabinet is added in the children?

components/molecules/InterstitialScreen/InterstitialScreen.story.jsx
```diff
@@ -15,15 +12,10 @@ stories.addParameters({
...
 });
 
-const stagingStyles = {
-  minHeight: '100vh',
-  paddingTop: '2rem',
-};
-
-const defaultProps = (steps) => ({
+const defaultProps = (actionBarOptions) => ({
+  isVerticallyAligned: boolean('isVerticallyAligned', false),
```
RohitRajendran, line 16: I don't see this as a prop?

components/molecules/InterstitialStepScreen/InterstitialStepScreen.scss
JamesIves: This is specific to the step box component as it needs additional padding for the bottom button. I think we should be agnostic in the default interstitial screen component.

components/molecules/InterstitialScreen/InterstitialScreen.story.jsx
JamesIves: Removed, leftover code.

components/molecules/InterstitialScreen/InterstitialScreen.scss
JamesIves: That's fair. I'm ok with moving it.

PR #1217, opened 2020-04-27
Adds the "Save Icon"

matthewheck:


components/atoms/icons/SaveIcon/SaveIcon.tsx
```diff
@@ -0,0 +1,59 @@
...
+    xmlns="http://www.w3.org/2000/svg"
+    style={
+      /** Workaround svg height and width attributes not supporting rems in Firefox and IE by passing it through style */
+      {
+        width,
+        height,
+        ...style,
+      }
+    }
+    viewBox="0 0 24 24"
```
JamesIves, line 36: Can you add an `aria-labelledby` prop here?

PR #1219, opened 2020-04-27
[APP-5044] SimpleSummary TypeScript

JamesIves:
Converts the `SimpleSummary` card to Typescript.

* There should be no changes.

N/A.

Had to fix up a few references to components that use this one in order to get it to compile correctly. 

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5044)


components/molecules/CardSummaries/SimpleSummary/SimpleSummary.tsx
```diff
@@ -1,13 +1,37 @@
...
+  shortTitle: React.ReactNode;
+} & (SimpleSummaryWithTo | SimpleSummaryWithEdit);
+
+const SimpleSummary: FC<SimpleSummaryProps> = ({
+  shortTitle,
+  editCardText,
+  answers,
+  ...props
+}) => {
+  const showEditButton = Boolean(props);
```
dylanbyars, line 31: This will always be `true`, won't it? 
```Boolean({}) === true```

Comment (JamesIves) [reply components/molecules/CardSummaries/SimpleSummary/SimpleSummary.tsx]: Just pushed up a change, I was missing something here

File: components/molecules/CardSummaries/SimpleSummary/SimpleSummary.tsx [hunk 2]
```diff
@@ -42,13 +42,13 @@ const SimpleSummary: FC<SimpleSummaryProps> = ({
         {showEditButton && (
           <div className="uic--col-6 uic--col-sm-1 uic--order-sm-last">
             <div className="uic--d-flex uic--justify-content-end">
-              {propsWithTo && propsWithTo.to && (
+              {showEditButton && propsWithTo && propsWithTo.to && (
```
dylanbyars, line 45: Why is `showEditButton` needed here? It's already true if this block renders.

JamesIves: Oops!

dylanbyars:
One last thing: convert SimpleSummary's tests to use testing-library. I'm planning on doing the same with all the other ts-ification stories.

PR #1220, opened 2020-04-28
adds more icons commonly used in buttons

matthewheck:
Adds the Add (plus sign) icon and the "trades" icon.

### DIFF
--- a/components/atoms/icons/AddIcon/AddIcon.tsx
+++ b/components/atoms/icons/AddIcon/AddIcon.tsx
@@ -1,0 +1,59 @@
+/** @module AddIcon */
+import React, {FC, memo} from 'react';
+
+type AddIconProps = {
+  /** Title for the svg. */
+  title?: string;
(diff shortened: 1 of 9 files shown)


components/atoms/icons/AddIcon/AddIcon.tsx
```diff
@@ -0,0 +1,59 @@
+/** @module AddIcon */
+import React, {FC, memo} from 'react';
+
+type AddIconProps = {
+  /** Title for the svg. */
+  title?: string;
```
RohitRajendran, line 6: Does title need to be a mandatory prop if it's expected by the aria label?

matthewheck: you don't have to pass it through, but it has a default.

RohitRajendran: Ah I missed that it had a default! Nevermind!

PR #1222, opened 2020-04-29
[APP-5042] Convert MessageCard to TS

JamesIves:
Converts the `MessageCard` component to use TypeScript. Also moved the `SimpleSummary` tests to be a `tsx` file as I missed that in the last PR.

* Observe the typescriptness in all its glory

N/A

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5042)


components/molecules/CardSummaries/SimpleSummary/SimpleSummary.tsx
```diff
@@ -5,12 +5,12 @@ import '../CardSummaries.scss';
 
 type SimpleSummaryWithTo = {
   /** The URL that the user should be directed to when edit is clicked, used instead of `editCard` for Hogwarts Express. */
-  to: string;
+  to?: string;
```
dylanbyars, line 8: Shouldn't these both not be optional?

JamesIves: They are technically both optional because we decide to hide or show the button if neither of them are present

components/organisms-simple/MessageCard/MessageCard.spec.tsx
```diff
@@ -5,13 +5,17 @@ import MessageCard from './MessageCard';
...
-    onSubmit: () => true,
+    onSubmit: (): undefined => undefined,
     name: 'test',
     title: 'Title',
     shortTitle: 'Q',
   };
 
-  const comp = mount(<MessageCard {...props}>Content</MessageCard>);
+  const comp = mount(
+    <MessageCard editCard={(() => null) as React.MouseEventHandler} {...props}>
```
RohitRajendran, line 15: Minor but why not just add the edit card value as part of the props object?

components/organisms-simple/MessageCard/MessageCard.tsx
```diff
@@ -1,14 +1,54 @@
...
+  /** To display something after the Submit button. */
+  afterButton?: React.ReactNode;
+  /** To display something before the Submit button. */
+  beforeButton?: React.ReactNode;
+  /** Changes the text in the Submit button. */
+  buttonText?: string;
+  /** Additional class names to apply to the button. */
+  className?: string;
+  /** The cards description. */
+  description?: React.ReactNode | string;
```
RohitRajendran, line 24:
```suggestion
  description?: React.ReactNode;
```
I don't think you actually need both reactnode and string

components/organisms-simple/MessageCard/MessageCard.tsx
```diff
@@ -1,14 +1,54 @@
...
+  isFetching?: boolean;
+  /** Shows a loading indicator on the button for actions after the button is clicked. */
+  loading?: boolean;
+  /** The handler to fire when the Submit button is clicked. */
+  onSubmit: () => undefined;
+  /** The index of this card in the flow, used for animation purposes */
+  stepIndex?: number;
+  /** The title of the card. */
+  title: React.ReactNode;
+};
```
RohitRajendran, line 49: I think it would be better to Pick<> and combine with the shared types from the CardShell rather than list all the types again

JamesIves: Waiting on this to be completed by @dylanbyars - I'll put this on hold until those are done

components/organisms-simple/MessageCard/MessageCard.spec.tsx
JamesIves: I wasn't able to cast the type when defining it through the object. Not really sure why, we do the same thing in the `Button` component I assume for the same reason.

components/organisms-simple/MessageCard/MessageCard.tsx
matthewheck: why ya picking on @dylanbyars ?

PR #1225, opened 2020-05-01
ActionBar Component

JamesIves:
Adds the `ActionBar` component. This mostly just abstracts the existing one from McG, but also adds the ability to have a back button. 

* It should behave the same in McG.
* It's generic enough to be used on top of whatever screen needs it.


components/molecules/ActionBar/ActionBar.tsx
```diff
@@ -0,0 +1,68 @@
...
+  /* Additional class names to apply to the container. */
+  className?: string;
+  /** Additional style properties to apply to the container. */
+  style?: React.CSSProperties;
+  /** The handler which should run when the close button is pressed. */
+  onClose?: React.MouseEventHandler;
+  /** The handler which should run when the back button is pressed. */
+  onBack?: React.MouseEventHandler;
+  /** The title of the screen that appears in the middle of the action bar. */
+  title?: string;
```
RohitRajendran, line 18: Should the title really be optional?

JamesIves: I think so, it works with or without it, so I think I'd prefer to keep it optional. The interstitials where these appear don't always have a title so it makes sense.

I just noticed though that the `h1` element remains even if it's not defined, I'll push an update so that's replaced with a `span` tag like the other items on the bar.

PR #1227, opened 2020-05-04
[APP-5098] Tile Component

JamesIves:
Implements the initial base `Tile` component. It supports `children` and a `footerContent` prop.

* Take a look at it, it's basically just a box with a border.
* Because the idea is for this component to support any content type, there's no need for it to enforce any sort of styling. You should assign a class to the item before you pass it in to this component. 

Screen Shot 2020-05-04 at 3 21 17 PM
Screen Shot 2020-05-04 at 3 21 25 PM

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5098)


components/atoms/Tile/Tile.tsx
```diff
@@ -0,0 +1,34 @@
+/** @module Tile */
+import classNames from 'classnames';
+import React, {FC} from 'react';
+import './Tile.scss';
+
+type TileProps = {
```
RohitRajendran, line 6: What about the blue color version?

JamesIves: Good call, I forgot about that. Give me a minute!

components/atoms/Tile/Tile.scss
```diff
@@ -0,0 +1,28 @@
+.uic--tile {
+  background-color: $white;
+
+  &--dark {
+    background-color: $midnight;
+
```
RohitRajendran, line 6:
```suggestion
    background-color: $midnight;
    color: $white
```
I think we should make the text white by default when dark

JamesIves: I'm ok with that

PR #1229, opened 2020-05-05
tsc Errors

JamesIves:
Resolves some issues with missing props and types being on the cards. I don't know if we can even do `yarn link` anymore, it seems to break a huge amount of things if it's done and built. 

Anyway, I think these changes should work.


components/organisms-simple/MessageCard/MessageCard.tsx
```diff
@@ -46,6 +46,8 @@ type MessageCardProps = {
   stepIndex?: number;
   /** The title of the card. */
   title: React.ReactNode;
+  /** A shorter version of the card title. */
+  shortTitle?: React.ReactNode;
```
RohitRajendran, line 50: Matthew pointed out that short title isn't actually used in messagecard. We should just make that fix in client

matthewheck: already have it ready.

PR #1231, opened 2020-05-05
[APP-5041] typescript QuestionCard component

dylanbyars:
### Description
Converted the `QuestionCard` component and its spec file to typescript.

### Testing Instructions
Check that there are no visual or behavior changes in the storybook

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5041)


components/organisms-complex/InputCard/InputCard.tsx
```diff
@@ -73,19 +75,15 @@ type InputCardConfig = {
   sanitize?: boolean;
 };
 
-type InputCardProps = typeof QuestionCard.propTypes & {
+type InputCardProps = QuestionCardProps & {
```
RohitRajendran, line 78: There are some props from QuestionCardProps that don't apply such as summary. Might actually just be summary.

components/organisms-simple/QuestionCard/QuestionCard.tsx
```diff
@@ -75,7 +139,7 @@ const QuestionCard = ({
       afterButton={
         <Fragment>
           {afterButton}
-          {!isLatestCard && (
```
RohitRajendran, line 78: We should just make `cancelChanges` required

components/organisms-complex/InputCard/InputCard.tsx
dylanbyars: But `InputCard` passes all of its props into `QuestionCard` so shouldn't it also expect every possible prop for `QuestionCard`?

RohitRajendran: But because `{...props}` is before `summary`, even if you pass in summary as part of props, it won't be used.

dylanbyars: Ahh yes I see what you mean. I'll update the props. Thanks!

PR #1240, opened 2020-05-07
App 5041 question cards ts 2

dylanbyars:
### Description
The `cancelChanges` props in the `QuestionCard` component _should_ be optional since there are situations in client where there is no cancel changes button rendered on a `QuestionCard` (see InvestmentPlanInfo.cards.tsx).

The `SliderCard` also needed a small adjustment to its props.

Finally, there's a stand alone change that @angelojcruzjr asked me to include here to fix the home button in storybook.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5041)


RohitRajendran:
After looking into the `cancelChanges` issue further, I think the issue might actually stem from `InvestmentPlanInfoCardsProps` missing `cancelChanges`. I think I was wrong about this needing a component library change

components/organisms-complex/SliderCard/SliderCard.tsx
```diff
@@ -24,15 +24,15 @@ type SliderCardConfig = {
...
+type SliderCardProps = Omit<QuestionCardProps, 'summary' | 'children'> & {
   /** Values to pass into select button */
   config: SliderCardConfig;
   /** The card answer(s) to show in the collapsed summary. If not provided, will use the label of the seleted value */
   answers?: string[];
   /** Handler called to edit the card, only necessary for McGonagall. */
   editCard: React.MouseEventHandler;
   /** The handler to fire when a change happens. */
-  onChange: typeof PureSlider.propTypes.onChange;
+  onChange: () => void;
```
RohitRajendran, line 35: I noticed on change and editCard are both part of QuestionCard props, do we need to declare it again here?

dylanbyars: Yes because `onChange` and `editCard` are both optional in `QuestionCard`. I explicitly set them to non-optional when other components that require either of those props extend `QuestionCardProps`.

PR #1244, opened 2020-05-11
App 5043 radio button card ts

dylanbyars:
### Description
* Modify some types in `components` to take advantage of event specific type definitions provided by react.
* Update `RadioButtonCard`'s test to use react-testing-library

### Testing Instructions
Ensure `RadioButtonCard` continues to work as expected.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5043)


components/organisms-complex/RadioButtonCard/RadioButtonCard.tsx
```diff
@@ -83,7 +83,7 @@ const RadioButtonCard: FC<RadioButtonCardProps> = ({
             ]
           }
           editCard={editCard}
-          shortTitle={shortTitle || title}
+          shortTitle={shortTitle}
```
JamesIves, line 86: Is this change made because `shortTitle` is required?

dylanbyars: Correct-o

dylanbyars: I checked all uses of `RadioButtonCard` in tsx files in client and they all have a `shortTitle` prop defined. Other uses can be checked/corrected as they're ts-ified.

components/organisms-complex/RadioButtonCard/RadioButtonCard.tsx
```diff
@@ -52,7 +52,7 @@ type RadioButtonCardProps = Omit<QuestionCardProps, 'summary'> & {
   /** The card answer(s) to show in the collapsed summary. If not provided, will use the label of the seleted value */
```
RohitRajendran, line 52: Instead of RadioButtonCardConfig, should that just be referencing the types from RadioButtons since that's what it's essentially referencing?

dylanbyars: It should but `RadioButtons` isn't typed yet. When it is, the types can be shared.

I'll improve the types defined here in the card so they match what `RadioButtons` will need exactly. Then converting `RadioButtons` will be a breeze.

PR #1245, opened 2020-05-11
[APP-5108] Double Donut Component

JamesIves:
Adds the double donut component. The API has been changed slightly during the port to make it more streamlined. It has all been documented in the docs file. 

* Mess around with the props, it should all make sense.
* You should be able to select each donut slice to expand the categories.
* Clicking the legend does not do anything. I'm unsure if this was the case before but that functionality doesn't seem present in production, and therefore it's not present in the port.
* You can now resize the double donut using the `height` and `width` props. There's an example in the story file.
* Click the very center of the donut 6 times for a surprise!

Screen Shot 2020-05-11 at 4 30 34 PM
Screen Shot 2020-05-11 at 4 31 02 PM

I'm doing my best to bring the test coverage up on this component. There's a _lot_ going on here but I don't want it to block any additional progress. 

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5108)


CalMlynarczyk:
I can't get this to compile.

JamesIves:
@CalMlynarczyk Compile error should be resolved now.

package.json
```diff
@@ -56,6 +56,10 @@
     "@storybook/react": "5.3.18",
     "@testing-library/react": "9.5.0",
     "@types/classnames": "2.2.10",
+    "@types/d3-interpolate": "^1.3.1",
```
RohitRajendran, line 59: Use exact versions for all packages

components/molecules/DoubleDonut/DoubleDonut.tsx
```diff
@@ -0,0 +1,374 @@
...
+            categoryKeys={categoryKeys}
+            categories={categories}
+          />
+        )}
+      </div>
+    );
+  }
+}
+
+DoubleDonut.defaultProps = {
```
RohitRajendran, line 369: You already declare this above

PR #1246, opened 2020-05-11
[APP-5080] Create chance of success display for no percent

CalMlynarczyk:
### Description
Differentiate between when we are waiting on a chance of success to be provided and when there is no chance of success value (ex. invest-only plans). I also found the text size was incorrect when in "compact" mode and "TBD" or "N/A".

### Testing Instructions
I added an additional story called **missing percent** to demonstrate the different modes.

### Comments
This change is not backwards compatible and will require updates to calling sites to maintain current behavior.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5080)


components/atoms/ChanceOfSuccess/ChanceOfSuccess.story.jsx
```diff
@@ -80,3 +80,13 @@ stories.add(
   ),
   {percy: {skip: true}},
 );
+
+stories.add(
+  'missing percent',
+  () => (
+    <ChanceOfSuccess
+      {...defaultProps({draft: false, running: false, compact: false})}
+    />
```
angelojcruzjr, line 89: Can we add a background color to this story and the others, similar to how we do in the markdown file for this component? It took me a bit to realize the text was there with it being white.

```suggestion
    <div style={{backgroundColor: 'black'}}>
      <ChanceOfSuccess
        {...defaultProps({draft: false, running: false, compact: false})}
      />
   </div>
```

JamesIves: There's a background toggle at the top of the story view. If you add a fixed background it will eliminate the ability to test it on other background colors. I think it makes sense in the docs file as such a thing doesn't exist there.

angelojcruzjr: Ah I was not aware of this configuration. I just did some research and it looks like we add the `backgrounds` param in `config.js` and set the default to white. Would it be okay to do something like this just for this set of stories (and any others like it in the future):

```javascript
...
import {colors} from '~constants/js/colors';

const stories = storiesOf('Atoms/ChanceOfSuccess', module);

stories.addParameters({
  docs: {
    page: ChanceOfSuccessReadme,
  },
  backgrounds: [
    {name: 'White', value: colors['white']},
    {name: 'Parchment', value: colors['parchment']},
    // Or any other color that's not white
    {name: 'Royal', value: colors['royal'], default: true},
    {name: 'Midnight', value: colors['midnight']},
    {name: 'Middle Purple', value: colors['midnight-purple']},
  ]
});
...
```

angelojcruzjr: Also, looking at these colors configs I actually think there's a bug - I think `midnight-purple` is meant to be `middle-purple` as there's no `midnight-purple` in the `colors.js` constants.

JamesIves: That approach makes the most sense to me. Perhaps we could make the array of objects shared.

CalMlynarczyk:
Are then any functional changes that need done for this PR? If everything else is good then I am going to merge this. This background color issue is a StoryBook thing and has very little to do with the ticket I'm working on.

PR #1248, opened 2020-05-12
Add Aria Label for Slider Handle

maggie9595:
### Description
Fix accessibility violation for Slider component by adding the aria label for slider handle.

### Testing Instructions
Check that the Slider handle has the aria label when it renders and defaults to `slider handle` if the prop is undefined.

### Screenshots
image

image


components/atoms/Slider/Slider.jsx
```diff
@@ -19,6 +19,7 @@ const Slider = ({
   tooltipStickyVariant,
   className,
   required,
+  ariaLabelForHandle = 'slider handle',
```
JamesIves, line 22: You should use `Slider.defaultProps = { ariaLabelForHandle: 'slider handle' }` instead of defaulting the props like this.

PR #1252, opened 2020-05-14
[APP-5099] Pagination Tile

jamaalalleyne-c1:
Created the Pagination Tile Component. This allows different views to be paginated through within a tile.

Navigate to Paginated Tiles story in storybook and verify the layout and behaviour of each story.

- [APP-5099](https://unitedincome.atlassian.net/browse/app-5099)

### DIFF
--- a/components/atoms/PaginationControls/PaginationControls.tsx
+++ b/components/atoms/PaginationControls/PaginationControls.tsx
@@ -4,7 +4,7 @@ import React, {FC} from 'react';
 import './PaginationControls.scss';
 import Button from '../Button/Button';
 import CaretIcon from '../icons/CaretIcon/CaretIcon';
-import {colors} from '~constants/js/colors';
+import {colors} from '../../../constants/js/colors';
 
 type PaginationControlsProps = {
   /**Disables/Enables the previous button. */
@@ -46,7 +46,7 @@ const PaginationControls: FC<PaginationControlsProps> = ({
         onClick={previous}
       >
         <CaretIcon
-          direction={'left'}
+          direction="left"
           fill={previousDisabled ? colors['white-30'] : colors.white}
         />
       </Button>
@@ -58,7 +58,7 @@ const PaginationControls: FC<PaginationControlsProps> = ({
         onClick={next}
       >
         <CaretIcon
-          direction={'right'}
+          direction="right"
           fill={nextDisabled ? colors['white-30'] : colors.white}
         />
       </Button>
--- a/components/atoms/PaginationControls/PaginationControls.tsx
+++ b/components/atoms/PaginationControls/PaginationControls.tsx
@@ -1,0 +1,69 @@
+/** @module PaginationControls */
+import classNames from 'classnames';
+import React, {FC} from 'react';
+import './PaginationControls.scss';
+import Button from '../Button/Button';
+import CaretIcon from '../icons/CaretIcon/CaretIcon';
+import {colors} from '~constants/js/colors';
+
+type PaginationControlsProps = {
+  /**Disables/Enables the previous button. */
+  previousDisabled: boolean;
+  /**Disables/Enables the next button. */
+  nextDisabled: boolean;
+  /**Function that navigates to the previous page */
+  previous: () => void;
+  /** Function that navigates to next page.*/
+  next: () => void;
+  /* Additional class names to apply to the container. */
+  className?: string;
+  /** Additional style properties to apply to the container. */
+  style?: React.CSSProperties;
+};
+
+const PaginationControls: FC<PaginationControlsProps> = ({
+  previousDisabled,
+  nextDisabled,
+  previous,
+  next,
+  className,
+  style,
+}) => {
+  const containerClasses = classNames(
+    {
+      'uic--pagination-controls': true,
+    },
+    className,
+  );
+
+  return (
+    <div className={containerClasses} style={style}>
+      <Button
+        variant="secondary"
+        light
+        disabled={previousDisabled}
+        className="uic--pagination-controls__pagination-button uic--h-100"
+        onClick={previous}
+      >
+        <CaretIcon
+          direction={'left'}
+          fill={previousDisabled ? colors['white-30'] : colors.white}
+        />
+      </Button>
+      <Button
+        variant="secondary"
+        light
+        disabled={nextDisabled}
+        className="uic--pagination-controls__pagination-button uic--h-100"
+        onClick={next}
+      >
+        <CaretIcon
+          direction={'right'}
+          fill={nextDisabled ? colors['white-30'] : colors.white}
+        />
+      </Button>
+    </div>
+  );
+};
+
+export default PaginationControls;
--- a/components/molecules/PaginatedTile/PaginatedTile.mdx
+++ b/components/molecules/PaginatedTile/PaginatedTile.mdx
@@ -36,7 +36,7 @@ import {PaginatedTile} from '@unitedincome/components';
         <Button style={{textAlign: 'left'}} variant="link" to="#">
           View Montezuma the Cat
         </Button>
-      )
+      ),
     }}
   />
 </Preview>
--- a/components/molecules/PaginatedTile/PaginatedTile.mdx
+++ b/components/molecules/PaginatedTile/PaginatedTile.mdx
@@ -21,88 +21,14 @@ import {PaginatedTile} from '@unitedincome/components';
 
 <Preview>
   <PaginatedTile
-    items={[
-      'Monday',
-      'Tuesday',
-      'Wednesday',
-      'Thursday',
-      'Friday',
-      'Saturday',
-      'Sunday',
-    ]}
-    listTemplate={{
-      itemsPerPage: 3,
-      Template: ({paginatedItems}) => {
-        return (
-          <ul>
-            {paginatedItems.map((item) => (
-              <li>{item}</li>
-            ))}
-          </ul>
-        );
-      },
-    }}
-    tileProps={{
-      footerContent: (
-        <Button style={{textAlign: 'left'}} variant="link" to="#">
-          View Montezuma the Cat
-        </Button>
-      ),
-    }}
-  />
-</Preview>
-
-```javascript
-  <PaginatedTile
-    items={[
-    'Monday',
-    'Tuesday',
-    'Wednesday',
-    'Thursday',
-    'Friday',
-    'Saturday',
-    'Sunday',
-  ]}
-  listTemplate={{
-      itemsPerPage: 3,
-      Template: ({paginatedItems}) => {
-      return (
-        <ul>
-          {paginatedItems.map((item) => (
-            <li>{item}</li>
-          ))}
-        </ul>
-      );
-    },
-  }}
-  tileProps={{
-    footerContent:
-      <Button style={{textAlign: 'left'}} variant="link" to="#">
-        View Montezuma the Cat
-      </Button>
-  }}
-  />
-  />
-```
-
-### No ListTemplate
-
-The items rendered in the pagination tile can be different views as well.
-In that case, embed the list of views(markup) in the `items` property and make sure to omit the `ListTemplate` property.
-
-<Preview>
-  <PaginatedTile
-    items={[
-      <React.Fragment>
-        <p>
-          This is a pagination example where the pages don't follow the same
-          markup/template.
-        </p>
-      </React.Fragment>,
-      <ul>
-        {['Monday', 'Tuesday'].map((item) => (
-          <li>{item}</li>
-        ))}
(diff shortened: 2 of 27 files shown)


components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,164 @@
...
+  className: 'uic--default-placeholder',
+  itemsPerPage: 1,
+};
+
+type PageFooterProps = {
+  footerContent?: React.ReactNode;
+  pagination: Pagination;
+  previous: () => void;
+  next: () => void;
+};
```
JamesIves, line 102: Can you separate these two with a line break?

components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,164 @@
...
+  return fn({...props});
+};
+
+PaginatedTile.defaultProps = {
+  className: 'uic--default-placeholder',
+  itemsPerPage: 1,
+};
+
+type PageFooterProps = {
+  footerContent?: React.ReactNode;
```
JamesIves, line 98: Missing comments on the prop types.

components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,164 @@
...
+  previous: () => void;
+  next: () => void;
+};
+const PageFooter: FC<PageFooterProps> = ({
+  footerContent,
+  pagination,
+  previous,
+  next,
+}) => {
+  return (
```
JamesIves, line 109: This could use an implicit return

components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,164 @@
...
+            onClick={next}
+          />
+        </div>
+      )}
+    </div>
+  );
+};
+
+type PaginationButtonProps = {
+  disabled: boolean;
```
JamesIves, line 139: Missing prop type comments

components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,164 @@
...
+  onClick: () => void;
+  direction: 'right' | 'left';
+};
+const PaginationButton: FC<PaginationButtonProps> = ({
```
JamesIves, line 143: Can you add a line break between the type and the const? Makes it easier to read

components/molecules/PaginatedTile/PaginatedTile.mdx
```diff
@@ -0,0 +1,213 @@
...
+/>
+```
+
+## Props 🔧
+
+<Props of={PaginatedTile} />
+
+## Related ✨
+
+- [Tile](/?path=/docs/atoms-tile--default)
```
JamesIves, line 213: Can you add `TabbedTile` here? And can you add a reference to `PaginatedTile` in `Tile` and `TabbedTile`?

components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,165 @@
...
+  /** Number of items to be displayed per page. This is used in conjuction with the ListTemplate property. */
+  itemsPerPage: number;
+  /** Determines if the navigation tile should be dark or not. */
+  isDark?: boolean;
+};
+
+const PaginatedTile = function <T>(
+  props: PaginatedTileProps<T>,
+): ReactElement | null {
+  const fn: FC<PaginatedTileProps<T>> = ({
```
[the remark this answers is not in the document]
jamaalalleyne-c1: i resorted to this because the PaginatedTileProps is a generic type and I couldn't find a way to declare FC with a generic props type other than if it was wrapped in another function

Essentially I couldn't find a way to resolve the generic type with the below structure:

const PaginatedTile: FC<PaginatedTileProps<T>> = ({...props}) => {}

Any suggestions as to how to do this?

components/molecules/PaginatedTile/PaginatedTile.tsx
jamaalalleyne-c1: I am assuming you mean this

```
const PaginationButton: 
FC<PaginationButtonProps>
```
the code formatting is reverting that and placing it on a single line.

components/molecules/PaginatedTile/PaginatedTile.tsx
jamaalalleyne-c1: This seems to conflict with an earlier comment made above

`I don't think you need to include JSX.Element as the return type if you use FC. There's a few places this occurs.`

Which should I follow?

components/molecules/PaginatedTile/PaginatedTile.scss
```diff
@@ -0,0 +1,35 @@
+.uic--paginated-tile {
+  .uic--tile__footer {
+    height: 5rem;
+    padding-bottom: 0;
```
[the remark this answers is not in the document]
jamaalalleyne-c1: I tried this `padding: 0 0 0 inherit` and a few other variations but they were all seen as invalid. any suggestions?

components/molecules/PaginatedTile/PaginatedTile.tsx
JamesIves: Sorry I mean between line 142 and 143

jamaalalleyne-c1:
> Looks good, couple of changes. Also the following items:
> 
> * The padding on the footer looks incorrect, the buttons in Storybook look squashed vs the comps. Also the footer padding is inconsistent with what is displayed in the `Tile` and `TabbedTile` components.
> 
> Screen Shot 2020-05-15 at 11 51 12 AM
> 
> Screen Shot 2020-05-15 at 11 52 19 AM

These two issues could be related to the same thing. 

In order to have the buttons stretch to the top edges of the footer, I set the top and bottom padding of the footer to 0 and then stretch the buttons to their flex container.

After doing that the footer section appears squashed so I gave it a height of 5rem which should explain the added rules to PaginatedTile.scss

```.uic--tile__footer {
    height: 5rem;
    padding-bottom: 0;
    padding-right: 0;
    padding-top: 0;
  }```

After doing this the buttons didn't appear as squashed(see screenshot):

<img width="1457" alt="Screen Shot 2020-05-15 at 3 43 12 PM" src="https://user-images.githubusercontent.com/61422320/82090520-bc4a5a80-96c3-11ea-9881-5afaef824229.png">

,but nonetheless I figured there could be a better way to do this.

Is there a way to preserve the footer sections padding(top and bottom) and still have the buttons stretch to the container's edges?

File: components/molecules/PaginatedTile/PaginatedTile.tsx [hunk 9]
```diff
@@ -0,0 +1,165 @@
...
+    <Button
+      variant="secondary"
+      light
+      disabled={disabled}
+      className="uic--paginated-tile__pagination-button"
+      onClick={onClick}
+    >
+      <CaretIcon
+        direction={direction}
+        fill={disabled ? 'rgb(122,84,199)' : colors.white}
```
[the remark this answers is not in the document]
JamesIves: Taking another look at this again, it looks like the disabled state uses `$white` with an opacity of `.3`. You should be able to replace the RGB value entirely with `white-30`.

components/molecules/PaginatedTile/PaginatedTile.tsx
JamesIves: By implicit return I mean you can remove the `return statement, and follow a pattern like this:

```javascript
const stuff: FC<StuffProps> = () => ( <div /> )
```

components/molecules/PaginatedTile/PaginatedTile.tsx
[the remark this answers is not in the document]
JamesIves: I'm confused about why it needs to be a generic type, couldn't you just force it to be either a `React.ReactNode` or a `string` and then make the function read something like this?

```javascript
type PaginatedTileProps = {
  /** Items to be rendered in the ListTemplate property or individual views to be rendered as pages.*/
  items: React.ReactNode | string[];
  /** Template used to render a list of items.*/
  listTemplate?: {
    /** Number of items to be displayed per page.*/
    itemsPerPage: number;
    /** Template markup.*/
    Template: React.ComponentType<{paginatedItems: React.ReactNode | string[]}>;
  };
  /**Additional Tile Properties*/
  tileProps?: TileProps;
};

const PaginatedTile: FC<PaginatedTileProps> = ({items, listTemplate, tileProps}) => {
    const [currentPage, setCurrentPage] = useState(0);

      {
        'uic--paginated-tile': true,
      },
      tileProps?.className,
    );

  // ...
```

JamesIves:
Still a few styling issues, the height
Screen Shot 2020-05-18 at 9 52 48 AM
 of the footer is inconsistent with the comps:

Screen Shot 2020-05-18 at 9 50 32 AM

components/molecules/PaginatedTile/PaginatedTile.tsx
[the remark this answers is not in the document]
jamaalalleyne-c1: With the discussion we had on slack, it no longer needs to be a generic type since items will be a list of views(markup), so React.ReactNode[] will suffice.

jamaalalleyne-c1:
> Still a few styling issues, the height of the footer is inconsistent with the comps:
> 
> Screen Shot 2020-05-18 at 9 52 48 AM
> 
> Screen Shot 2020-05-18 at 9 50 32 AM

This seems to be coming from the padding in the Tile component, for the footer component. Will adjust it there.

components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,56 @@
...
+import React, {FC, useState} from 'react';
+import './PaginatedTile.scss';
+import Tile, {TileProps} from '~components/atoms/Tile/Tile';
+import PaginationControls from '~components/atoms/PaginationControls/PaginationControls';
+
+type PaginatedTileProps = {
+  /** List of pages to be rendered in the Tile.*/
+  pages: React.ReactNode[];
+  /**Additional Tile Properties*/
+  tileProps?: TileProps;
```
JamesIves, line 12: I can go either way on this, but for the typing here would it be better to do the following?

```typescript
type PaginatedTileProps = {
  pages: React.ReactNode[];
} & TileProps;
```

This would prevent you form needing to reference things like `tileProps?.className` and it would simplify how these are passed in. I know the `TabbedTile` does something similar to this but technically it renders multiple tiles vs a single one. 

CC: @jamaalalleyne-c1 @RohitRajendran

components/molecules/PaginatedTile/PaginatedTile.tsx
```diff
@@ -0,0 +1,56 @@
+/** @module PaginatedTile */
+import classNames from 'classnames';
+import React, {FC, useState} from 'react';
+import './PaginatedTile.scss';
+import Tile, {TileProps} from '~components/atoms/Tile/Tile';
+import PaginationControls from '~components/atoms/PaginationControls/PaginationControls';
```
JamesIves, line 6: I worry that using module path aliasing may cause an issue when injected into our projects. See ticket APP-5119 for details. We never properly documented or made a call on this as a team. 

So now we may need to use absolute paths in TypeScript files until that ticket is resolved or at least looked at.

components/atoms/PaginationControls/PaginationControls.tsx
```diff
@@ -0,0 +1,69 @@
...
+      </Button>
+      <Button
+        variant="secondary"
+        light
+        disabled={nextDisabled}
+        className="uic--pagination-controls__pagination-button uic--h-100"
+        onClick={next}
+      >
+        <CaretIcon
+          direction={'right'}
```
JamesIves, line 61:
```suggestion
          direction="right"
```

No need to use brackets here as it's a string.

components/atoms/PaginationControls/PaginationControls.tsx
```diff
@@ -0,0 +1,69 @@
...
+    <div className={containerClasses} style={style}>
+      <Button
+        variant="secondary"
+        light
+        disabled={previousDisabled}
+        className="uic--pagination-controls__pagination-button uic--h-100"
+        onClick={previous}
+      >
+        <CaretIcon
+          direction={'left'}
```
JamesIves, line 49:
```suggestion
          direction="left"
```

components/molecules/PaginatedTile/PaginatedTile.tsx
RohitRajendran: My vote is for `{
  pages: React.ReactNode[];
} & TileProps;`

PR #1256, opened 2020-05-18
App 5072 mc gonagall ts

dylanbyars:
### Description
Added types to McGonagall component.

Highlights: 
* A new `McGonagallCardProps` type that can be used as the type for the the props argument of the `McGonagall` component's `card` prop (for rendering cards on steps).
* 

### Testing Instructions
* check `components` repo tests
* check  `client` repo tests (once integration is complete. See PR)

Updating the associated spec file to ts and converting it to use RTL was split out to another ticket.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5072)


package.json
```diff
@@ -100,7 +101,7 @@
     "postcss-prefixer": "2.1.2",
     "prettier": "2.0.4",
     "prop-types": "15.7.2",
-    "query-string": "=5.1.1",
+    "query-string": "6.12.1",
```
RohitRajendran, line 104: This is hardcoded to 5.1.1 in both client and component library. I want to say this was because it broke IE or something in future versions. Maybe @JamesIves remembers?

dylanbyars: I had a hunch it was set at a specific version for a reason. `query-string` bundles includes its own type definitions as of version 6.8.3. They don't keep types separated out, unfortunately as far as I can tell. If this version breaks something then I'll need to rethink the way we get those types.

matthewheck: I'm like 99% sure that query string v6 broke IE11. Let's keep it at 5 until we drop IE11 support on June 19.

components/organisms-simple/MessageCard/MessageCard.tsx
```diff
@@ -41,7 +41,10 @@ type MessageCardProps = {
   /** Shows a loading indicator on the button for actions after the button is clicked. */
   loading?: boolean;
   /** The handler to fire when the Submit button is clicked. */
-  onSubmit: (e: SyntheticEvent) => Promise<void>;
+  onSubmit: (
+    e: FormEvent,
+    outputDefaults?: Record<string, string | boolean>,
+  ) => Promise<void>;
```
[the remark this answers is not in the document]
davery-cof: I think we could run into trouble with that interface if and when we convert CardShell to typescript because CardShell expects an async function, and I wouldn't prefer that interface in general, but since we already do it all over the place and MessageCard passes onSubmit to CardShell only, I don't have any real issue with the return value being Promise<void> or void..

I guess the alternative would be to make sure that all the McGonagall functions are either async or return a promise or void intrinsically with something like this:

```
const promiseFn = (): Promise<void> => {
  return new Promise((resolve) => resolve());
};
```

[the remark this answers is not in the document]
dylanbyars: I'm not a fan of wrapping synchronous functions in promises to make a dependent thing's types happy. The real solution is to take a look at all of the components in the components library and define some common shareable types. I think a lot of the mess with the types in components now has to do with individual components getting typed in relative isolation. It was a necessary strategy when we started converting to ts and we can now make connections between similar types with the help of the exploded out view.

There should be 2 flavors of type for most organisms in components. One for when they're used as a standalone thing and one for when they're a part of a McG (or Hogwarts) flow. Taking this general stance will almost certainly help identify the many types that are shared between those use cases so we can focus on defining clearer types for props that work differently in the different contexts.

I think the `Promise<void> | void` return is the least intrusive way to make this work in the short term. I've added a tech debt item dedicated to exploring reusable types in component library for the upcoming innovation sprint.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -101,21 +243,30 @@ class McGonagall extends React.Component {
...
+
     let finalReached = false;
 
     do {
       if (
         latest.value !== currXState.value &&
         latest.value !== newCardHistory[0].name
       ) {
-        newCardHistory.unshift(this.getMatchingStep(this.props.steps, latest));
+        newCardHistory.unshift(this.getMatchingStep(this.props.steps, latest)!);
```
bryanjcof, line 269: Based on the signature of `getMatchingStep` I don't think this `!` is necessary.

```suggestion
        newCardHistory.unshift(this.getMatchingStep(this.props.steps, latest));
```

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -241,8 +408,11 @@ class McGonagall extends React.Component {
    * @param {object} state state machine
    * @returns {object} step
    */
-  getMatchingStep(allSteps, state) {
-    return allSteps.find((step) => state.matches(step.name));
+  getMatchingStep(
+    allSteps: Step[],
+    state: State<TContext, McGonagallEvent, StateSchema, any>,
+  ): Step {
+    return allSteps.find((step) => state.matches(step.name))!;
```
bryanjcof, line 415: Similar to my earlier comment about throwing an explicit error, based on the usage of the results from `getMatchingStep` I think something like the below will help developers catch bugs that should only happen in development and will give explicit errors for a thing that might otherwise be difficult to track down if it happens when deployed. `state.value` may not be the best piece of information to use in the error message but I think this suggestion conveys intent.

```suggestion
    const result = allSteps.find((step) => state.matches(step.name));
    if (result === null || typeof result === 'undefined') {
      throw new Error(`Unable to find step matching ${state.value}`);
    }
    return result;
```

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -125,18 +268,20 @@ class McGonagall extends React.Component {
       });
 
       // Stop right before final step
-      if (!this.isFinalStep(this.props.stateConfig.states, next.value)) {
+      if (
+        !this.isFinalStep(this.props.stateConfig.states, next.value as string)
```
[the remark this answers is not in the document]
bryanjcof: Based on this, I would consider creating an `extractValue` function that takes an xstate state and checks if `state.value` is a string and returning the value if it is a string and throwing an error if it is not a string. This would enforce the expectations encoded in types but also point at the places that need to change in the future if changes to support nested flows are made by highlighting (with runtime and compile time errors) where it fails. Because of the explicit casts this usage will only fail at runtime if a change for nested flows is made.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -194,10 +350,16 @@ class McGonagall extends React.Component {
     newState.actions.forEach((action) => {
       // If the action is executable, execute it
       action.exec &&
```
bryanjcof, line 352: I'd prefer if this checked the type of exec rather than only testing its existence. This gives a little more confidence (at runtime) that `exec` can be invoked. Though based on the type signature of `ActionObject` this is probably not necessary. This stems from my dislike of truthy checks in TypeScript but you're welcome to ignore this one!

```suggestion
      typeof action.exec === 'function' &&
```

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -194,10 +342,16 @@ class McGonagall extends React.Component {
...
-        });
+        action.exec(
+          this.state.currXState.context,
+          {
+            type: 'NEXT',
+            ...payload,
+            ...updatedPayloadValues,
+          },
+          // eslint-disable-next-line no-warning-comments
+          undefined as any, // this param is not optional and was not provided in the js version
```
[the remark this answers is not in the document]
bryanjcof: This is kind of yuck but I agree with the reasoning.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -213,6 +375,13 @@ class McGonagall extends React.Component {
       updatedCurrState = newState;
     }
 
+    const stepToNavigateTo =
+      this.activeCard !== this.state.currXState.value && !clearFuture
+        ? this.state.currXState.value
+        : newState.value;
+
+    this.navigateToStep(stepToNavigateTo as string);
```
bryanjcof, line 383: I think this is another place I would use the `extractValue` function I mentioned previously.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -221,18 +390,16 @@ class McGonagall extends React.Component {
...
+        cardHistory: [
+          this.getMatchingStep(this.props.steps, newState),
+          ...prevHistory,
+        ],
+      });
+    } else {
+      this.setState(updatedData);
     }
-
-    this.navigateToStep(
```
bryanjcof, line 230: Based on the implementation I don't think that reordering `navigateToStep` and the calls to `setState` will have an impact but I wanted to confirm this is an intended change to the order of calls.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -272,26 +446,25 @@ class McGonagall extends React.Component {
...
+  close(e: MouseEvent): void {
     e.preventDefault();
     if (this.props.onClose) {
       this.props.onClose();
     }
     // Ignoring this if since window.location can't be set in unit test
     /* istanbul ignore if  */
     if (this.props.exitLocation.startsWith('http') && isWindowDefined()) {
-      window.location = this.props.exitLocation;
+      window.location.pathname = this.props.exitLocation;
```
bryanjcof, line 467: The if here indicates this the `exitLocation` is not a path, should this be setting `href` instead?

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -146,9 +299,12 @@ class McGonagall extends React.Component {
    * @param {array<string>} outputs the expected outputs for the step
    * @param {object} outputDefaults default values for outputs to fallback to
    * @param {boolean} clearFuture whether it should clear values from future steps
-   * @returns {undefined}
    */
-  updateStateMachine(outputs, outputDefaults, clearFuture) {
+  updateStateMachine(
+    outputs: string[],
+    outputDefaults?: OutputDefaults,
+    clearFuture?: boolean,
```
bryanjcof, line 306: The only usage I found always passes `clearFuture`, do these parameters need to be optional? The signature could be changed so `outputDefaults` accepts `undefined` but all parameters are required. Maybe six of one half dozen of the other though.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -334,41 +507,51 @@ class McGonagall extends React.Component {
...
 
     // Reverts any component state changes and opens card
-    const editCard = () => {
+    const editCard = (): void => {
       // Check for changes in active step
       const currStepOutputs = this.props.steps.find(
         ({name}) => name === this.activeCard,
-      ).outputs;
-      const currStepStatePayload = pick(this.state, currStepOutputs);
+      )!.outputs;
```
bryanjcof, line 550: Rather than the non-null assertion here I'd prefer an explicit check and thrown error if a null value is found. It looks like all usages of `currStepOutputs` include the non-null assertion so it should amount to the same thing.

components/query-string.d.ts
```diff
@@ -0,0 +1,399 @@
+// Manually added types for query-string because the official types ship with a newer version of query-string that breaks in IE.
```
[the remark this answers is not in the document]
dylanbyars: I didn't do a detailed check. I just pulled the earliest version of the type definitions provided by `query-string` and threw them all in here. Does the component library need to support IE indefinitely? If so, then a more careful parsing of these types might be worthwhile.

components/templates/McGonagall/McGonagall.tsx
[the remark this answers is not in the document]
dylanbyars: Love this idea. I'll included it in a commit shortly.

components/query-string.d.ts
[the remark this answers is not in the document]
bryanjcof: Since we're dropping support for IE I'd say the component library does not need to have different compatibility tables than the application. If IE specific things are the differences then I'm fine with using the newer type definitions. I flagged it because I wouldn't want us to write against type definitions that are going to cause unexpected behavior at runtime because of the mismatch.

components/organisms-complex/SliderCard/SliderCard.tsx
```diff
@@ -32,7 +32,7 @@ type SliderCardProps = Omit<QuestionCardProps, 'summary' | 'children'> & {
   /** Handler called to edit the card, only necessary for McGonagall. */
   editCard: React.MouseEventHandler;
   /** The handler to fire when a change happens. */
-  onChange: () => void;
+  onChange: (fieldName: string, newValue: string) => void;
```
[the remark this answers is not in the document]
dylanbyars: It absolutely makes sense to organize the types for components that use McG but I tried to do as little of that as possible as a part of this work. I think the main prop getting me in trouble is the change handler so I might spend some time at least cleaning that bit up. The change you noted above in `RadioButtonCard` needs some attention...

components/organisms-simple/CompletionScreen/CompletionScreen.tsx
```diff
@@ -1,15 +1,27 @@
...
+import React, {ReactNode} from 'react';
 import './CompletionScreen.scss';
 
+type CompletionScreenProps = {
+  /** Action button or links to be displayed at the bottom of the screen */
+  actions: ReactNode;
+  /** Additional class names to apply */
+  className?: string;
+  /** Screen description */
+  description?: ReactNode;
```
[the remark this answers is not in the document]
dylanbyars: I did it so that the description prop was consistent with the other cards. I'm open to reverting this change if we think it's a problem.

components/organisms-simple/QuestionCard/QuestionCard.tsx
```diff
@@ -57,9 +57,16 @@ export type QuestionCardProps = {
...
   /** The handler to fire when a change happens. */
-  onChange?: React.ChangeEventHandler;
+  onChange?:
+    | React.ChangeEventHandler
+    | ((fieldName: string, newValue: string) => void);
   /** The handler to fire when the Submit button is clicked. */
-  onSubmit: React.FormEventHandler;
+  onSubmit:
+    | React.FormEventHandler
+    | ((
```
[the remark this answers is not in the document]
dylanbyars: I thought more about the `onChange` handler comments from earlier and I think I will do something similar for `onSubmit`. It's probably best to not punt on getting these types cleaner.

components/templates/McGonagall/McGonagall.story.tsx
```diff
@@ -188,7 +217,12 @@ const renderState = (props, state) => {
   );
 };
 
-const renderTotalRetirementAccountValue = (props, state) => {
+const renderTotalRetirementAccountValue = (
+  props: McGonagallCardProps,
+  state: {
```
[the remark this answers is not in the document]
dylanbyars: That's what this is doing but without the utility type. I don't know what this style of type reference is called but it returns the type I expect.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -1,27 +1,170 @@
...
+  /** Whether it should animate on mount */
+  animate: boolean;
+  /** Handler to cancel the changes. */
+  cancelChanges: () => void;
+  /** Whether making changes should clear out any changes made in steps after this. */
+  clearFuture?: boolean;
+  /** The card description. If given a raw string, the string is wrapped with a p tag. */
+  description: ReactNode;
+  /** Default values for a card */
+  defaultValues?: {}[];
```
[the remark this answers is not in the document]
dylanbyars: I'll fill in some more details.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -1,27 +1,170 @@
...
+    e: FormEvent,
+    outputDefaults?: OutputDefaults,
+  ) => void | Promise<void>;
+  /** Shorter title to be used with the card summary. */
+  shortTitle: ReactNode;
+  /** Shorter title to be used with the card summary. */
+  stepIndex: number;
+  /** The title of the card. */
+  title: ReactNode;
+  key: string;
```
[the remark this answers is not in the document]
dylanbyars: I don't know if it's _necessary_ but it was part of the object returned from `renderStep` so I accounted for it.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -1,27 +1,170 @@
...
+  /** Additional class names to apply */
+  className?: string;
+  /** Name of the flow */
+  name?: string;
+  /** Page to go to if user clicks the close button */
+  exitLocation: string;
+  /** Function to call on exit of the flow */
+  onClose?: () => void;
+  /** Location info from router */
+  location: {
```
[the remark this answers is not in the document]
dylanbyars: They totally could but then we'd be mandating that this component has to use that particular router. I left it fairly plain since this is supposed to be an open source library. But there are worse things than saying this component expects to integrate with a particular outside dependency. Honestly, I'd prefer to be explicit and include `react-router` as a named dependency here.

components/organisms-simple/CompletionScreen/CompletionScreen.tsx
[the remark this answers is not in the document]
bryanjcof: Based on the usage I think I like keeping it a string since it gets dropped into a `<p>` tag and putting a more general `ReactNode` in here could cause layout problems. Though I suppose that should be caught in visual QA. There's arguments both ways so I have no objection to leaving it as-is.

components/templates/McGonagall/McGonagall.story.tsx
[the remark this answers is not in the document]
bryanjcof: It does indeed do the same thing. I tend towards the `Pick` but if the tendency in this codebase is to use this style please ignore these comments. This is a tentatively held preference so even if not I don't have a strong argument in favor of `Pick`.

components/templates/McGonagall/McGonagall.tsx
dylanbyars: Excellent catch. I think you're right. Will fix

components/templates/McGonagall/McGonagall.tsx
dylanbyars: I reordered the args when I thought that only outputDefaults was optional. I'll revisit this.

components/templates/McGonagall/McGonagall.tsx
dylanbyars: I'm probably going to leave this one as is for now but I'd love to know more about your "dislike of truthy checks in TS".

bryanjcof: It mostly boils down to my feeling that the advantages of TypeScript come from requiring the developer to be explicit about their expectations regarding what is contained in a value and truthy checks ignore that. In this example you're relying on the implied type of the `action.exec` property to not change. If the type of `action.exec` does change this as written will fail at runtime and it won't necessarily be clear to the new developer about why. Encoding the expectation explicitly, if the type of `action.exec` were to be changed away from a function I believe `tsc` would at least warn and say "this can never be true".

components/organisms-simple/QuestionCard/QuestionCard.tsx
```diff
@@ -22,7 +22,7 @@ export type QuestionCardProps = {
   /** Handler to cancel the changes. */
   cancelChanges: () => void;
   /** HTML element that should appear within the card when not collapsed */
-  children: ReactNode;
+  children?: ReactNode;
```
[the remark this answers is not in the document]
dylanbyars: I can dig it.

components/global.d.ts
```diff
@@ -0,0 +1 @@
+declare module '*.mdx';
```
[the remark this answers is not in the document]
dylanbyars: TODO: add to `import-asset.d.ts`

components/templates/McGonagall/McGonagall.tsx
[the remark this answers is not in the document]
davery-cof: To add on to this conversation, this can also be typed as:

import { Location } from 'history';
location: Location

dylanbyars:
Alright, gang. I've responded to the feedback submitted when the PR was first open for review.

Big changes: 
1. no more non-null assertions. I'm either throwing an error or falling back to some other equivalent data structure (see the `editCard` function in the `renderStep` method.
2. added 1 RTL test to the existing spec file to test an error is thrown properly

package.json
```diff
@@ -61,6 +61,7 @@
     "@types/d3-shape": "1.3.2",
     "@types/d3-transition": "1.1.6",
     "@types/enzyme": "3.10.5",
+    "@types/lodash": "^4.14.150",
```
RohitRajendran, line 64: Can you make this the exact package version

components/templates/McGonagall/McGonagall.tsx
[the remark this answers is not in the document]
bryanjcof: It might be overkill, but maybe a `Values` type argument to `McGonagallCardProps` to use with `defaultValues` makes sense?

```
type McGonagallCardProps<V = {}> {
  defaultValues?: V[];
}
```

This would let the type for values be specified (or inferred) when provided as a prop.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -1,27 +1,163 @@
...
+  isFetching?: boolean;
+  /** Whether this is the furthest step. if this and isCollapsed is true, this will collapse to the incomplete summary/ */
+  isLatestCard: boolean;
+  /** The handler to fire when a change happens. */
+  onChange: (fieldName: string, newValue: string) => void;
+  /** The handler to fire when the Submit button is clicked. */
+  onSubmit: (e: FormEvent, outputDefaults: OutputDefaults) => void;
+  /** Shorter title to be used with the card summary. */
+  shortTitle: ReactNode;
+  /** Shorter title to be used with the card summary. */
```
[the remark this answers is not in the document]
bryanjcof: Looks like the documentation comment for `stepIndex` still needs to be updated.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -1,27 +1,161 @@
...
+  /** Whether it should clear all future values when this card is edited */
+  clearFuture?: boolean;
+  /** A function that takes in props and state and renders the card */
+  card: (cardProps: McGonagallCardProps, cardState: any) => ReactNode;
+  /** The names of all the outputs for this card, required for cards that update values */
+  outputs?: string[];
+  defaultValues?: {}[];
+};
+
+type McGonagallEvent = {type: 'NEXT'};
```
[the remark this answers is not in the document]
bryanjcof: Can any other properties exist on `McGonagallEvent` objects or are they always exactly the value `{type: 'NEXT'}`, basically what I'm asking is should this be intersected with something to better define the `McGonagallEvent` type? Something like `type McGonagallEvent = {type: 'NEXT'} & Record<string, string>` (I don't know that the Record definition is correct, only using as an example).

components/templates/McGonagall/McGonagall.tsx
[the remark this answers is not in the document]
bryanjcof: Could also be done as:

```
import {Location} from 'history';

  location: {
    pathname: Location['pathname'];
    search: Location['search'];
  }
```

That serves mostly as an indication that the two things are related since those types resolve to `string`.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -1,27 +1,172 @@
...
+export type McGonagallState<TContext = DefaultContext> = {
+  /** The name of the active step. */
+  activeCard: Step['name'];
+  /** array of completed steps */
+  cardHistory: Step[];
+  /** current xstate state */
+  currXState: State<TContext, McGonagallEvent, StateSchema, any>;
+  /** Used to switch animation styles */
+  hasMounted: boolean;
+} & DefaultContext; /** xstate context */
```
bryanjcof, line 154: Does  `McGonagallState` intersect `DefaultContext` or is `DefaultContext` being used as an alias for `Record<string, any>` here?

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -88,26 +235,54 @@ class McGonagall extends React.Component {
...
    */
-  get activeCard() {
-    return (
-      queryString.parse(this.props.location.search).step ||
-      this.state.currXState.value
-    );
+  get activeCard(): string {
+    const queryStringCard = queryString.parse(this.props.location.search).step;
+
+    if (typeof queryStringCard !== 'string') {
```
bryanjcof, line 241: Based on the return value, it looks like `null` or `undefined` are valid values for `queryStringCard`, so I suspect this should be slightly broader to check that `queryStringCard` exists and is not a string. Alternatively this could maybe use `Array.isArray` as the test for throwing an error.

components/templates/McGonagall/McGonagall.tsx
```diff
@@ -1,27 +1,172 @@
...
+  onClose?: () => void;
+  /** Location info from router */
+  location: {
+    /** Pathname */
+    pathname: string;
+    /** Query string */
+    search: string;
+  };
+  /** State machine config object that tells the state machine the different states and context. View https://xstate.js.org/docs/guides/machines.html#configuration for more details */
+  stateConfig: {
```
bryanjcof, line 116: Is this the same as parts of `MachineConfig` from `xstate`?

components/templates/McGonagall/McGonagall.tsx
bryanjcof: Re-upping this question.

components/templates/McGonagall/McGonagall.story.tsx
```diff
@@ -110,12 +120,21 @@ const stateOptions = {
...
+      ctx: DefaultContext,
+      eve: {supportDependents: StateContext['supportDependents']},
+    ): boolean => eve.supportDependents === 'No',
   },
 };
 
-const renderSupportDependents = (props, state) => {
+const renderSupportDependents = (
+  props: McGonagallCardProps,
+  state: {supportDependents: StateContext['supportDependents']},
```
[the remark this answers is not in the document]
dylanbyars: It's _really_ the entire state object from the `McG` component. I've typed it as `any` in the `Step` type 
``` ts
  card: (cardProps: McGonagallCardProps, cardState: any) => ReactNode;
```
The McG state includes the entire `StateContext` so I do think you're right. The type of state should be the full StateContext.

components/templates/McGonagall/McGonagall.tsx
dylanbyars: `McGState` does intersect with the initialContext of the state machine which has the same type as `DefaultContext`

components/templates/McGonagall/McGonagall.tsx
dylanbyars: The current step of a flow is always written to the query string. Anything other than a string means that one of those conditions isn't met.

components/templates/McGonagall/McGonagall.tsx
dylanbyars: No but it is a selection from `xstate`'s `StateNodeConfig`.

Are you thinking I should make that link more explicit?

components/templates/McGonagall/McGonagall.tsx
dylanbyars: This was not purposeful and I see no good reason to change the call order so I'll set it back to setting state before calling the `navigateToStep` function.

components/templates/McGonagall/McGonagall.tsx
bryanjcof: It was the direction I was headed yeah. I was thinking that if this is using the same types it makes sense to reference the xstate types.

PR #1259, opened 2020-05-20
[APP-5100] TableTile

jamaalalleyne-c1:
Added the TableTile component which renders a tabular structure within a tile.

Navigate to Table Tile story in storybook and verify the layout and behaviour of each story.

- [APP-5100](https://unitedincome.atlassian.net/browse/app-5100)


jamaalalleyne-c1:
> Looks good, just a few minor comments.
> 
> Also I'm noticing the buttons have their icons a little bit off center, seeing this in the paginated tile too.
> 
> Screen Shot 2020-05-21 at 12 14 30 PM

monique also mentioned this bug in her review of [APP-5099](https://unitedincome.atlassian.net/browse/APP-5099) so will address it there.

components/molecules/TableTile/TableTile.scss
```diff
@@ -0,0 +1,38 @@
...
+    text-transform: uppercase;
+  }
+
+  &__cell {
+    border-color: $light-gray;
+    border-style: solid;
+    border-width: 1px 0 0;
+    padding: 1.168rem 1.8rem 1.772rem;
+
+    &_even {
```
JamesIves, line 25: I think these should be `&--even` as they are modifiers (following the [BEM](http://getbem.com/naming/) naming conventions) 

`uic--table-tile` is the block, the `__cell` is the element, and the `&--even` is the modifier

components/molecules/TableTile/TableTile.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const containerClasses = classNames(
+    {
+      'uic--table-tile': true,
+    },
+    className,
+  );
+
+  const pages: React.ReactNode[] = [];
+
+  for (let page = 0; page < Math.ceil(items.length / itemsPerPage); page++) {
```
JamesIves, line 35: I don't have strong feelings on this either way, but I think you could make this slightly cleaner by using the following pattern:

```javascript
const pages: React.ReactNode[] = Array.from({length: Math.ceil(items.length / itemsPerPage)}, (v, i) => i).map(() => <div />)
```

I'll leave the choice up to you!

components/molecules/TableTile/TableTile.scss
jamaalalleyne-c1: sure I can change it to that. 

fyi, I was following the BEM methodology from the readme which leads [here](https://en.bem.info/methodology/quick-start/#modifier). I think this has the modifier preceded by a single underscore.

but I would agree that double hyphens is a clearer distinction.

components/molecules/TableTile/TableTile.tsx
jamaalalleyne-c1: I think I will decide not to go with this, my only reason being that the value field will be declared and not used, but thanks for the suggestion.

PR #1261, opened 2020-05-21
Adds isDark prop to the ChanceOfSuccess component

JamesIves:
Adds the option for the text in the chance of success meter to be dark. 

* Click `isDark`
* It will be dark!


JamesIves:
> if possible you will probably need to get the code cov up, but other than that lgtm.

Not really sure why it's deciding to flag this component as lower test cov, this change should is adequately represented in a test :(

PR #1271, opened 2020-05-29
COS Resizable

JamesIves:
Makes the COS wheel resizable via a prop value. 


components/atoms/ChanceOfSuccess/ChanceOfSuccess.jsx
```diff
@@ -242,7 +249,7 @@ const ChanceOfSuccess = (props) => {
             </g>
           )}
           <text
-            fill={isDark ? colors.midnight : colors.white}
+            fill={colors.white}
```
RohitRajendran, line 252: How come you're getting rid of the isDark?

components/atoms/ChanceOfSuccess/ChanceOfSuccess.jsx
```diff
@@ -323,8 +331,19 @@ ChanceOfSuccess.propTypes = {
   compact: PropTypes.bool,
   /** Adjusts the appearance of the component if the user is using Internet Explorer. */
   isIe: PropTypes.bool,
-  /** Renders the ChanceOfSuccess meter to appear more kindly on light backgrounds. */
+  /** Renders the ChanceOfSuccess meter to appear more kindly on light backgrounds. */	  /** The height of the meter. */
```
RohitRajendran: Typo

components/atoms/ChanceOfSuccess/ChanceOfSuccess.jsx
JamesIves: oopsie

components/atoms/ChanceOfSuccess/ChanceOfSuccess.jsx
JamesIves: also oopsie

components/atoms/ChanceOfSuccess/ChanceOfSuccess.jsx
```diff
@@ -323,8 +331,19 @@ ChanceOfSuccess.propTypes = {
...
+  /** The height of the meter. */
+  height: PropTypes.string,
+  /** The width of the meter. */
+  width: PropTypes.string,
+  /** Additional style properties to apply to the container. */
+  style: PropTypes.objectOf(PropTypes.string),
+};
+
+ChanceOfSuccess.defaultProps = {
+  height: '14.1rem',
```
RohitRajendran, line 345: Is the addition of these default props going to affect and current uses of this component?

JamesIves: I made the default props match what the default height used to be without a height/width value. It was hardcoded as `141px` in the stylesheet (which I've now removed)

PR #1273, opened 2020-05-29
[APP-5216] Upload Trades - Scaffold

davery-cof:
### Description
Changes to fix the interface for MessageCard, required for Upload Trade CSV

Part of four PR's:

API - https://github.com/UnitedIncome/api/pull/3157
Client - https://github.com/UnitedIncome/client/pull/4020
UI-common - https://github.com/UnitedIncome/ui-common/pull/490
Components - **This**

### Testing Instructions
See Client PR above for testing instructions.

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5126)


components/organisms-simple/MessageCard/MessageCard.tsx
```diff
@@ -41,7 +41,7 @@ type MessageCardProps = {
   /** Shows a loading indicator on the button for actions after the button is clicked. */
   loading?: boolean;
   /** The handler to fire when the Submit button is clicked. */
-  onSubmit: () => undefined;
+  onSubmit: (e: SyntheticEvent) => Promise<void>;
```
labichn-cof, line 44: This function is called with two arguments inside of the `CardShell` (https://github.com/UnitedIncome/components/blob/b2e812dab9bbe01758853f0ef167923b741ed629/components/molecules/CardShell/CardShell.jsx#L204). Should there be an optional second argument here? Or do we get away with omitting it because `outputDefaults` is guaranteed to be `undefined` in the `CardShell` of the `MessageCard`?

davery-cof: I think you are correct that since outputDefaults is not passed through nor is it required in the CardShell, then we are safe to not put it as a parameter of the function

PR #1276, opened 2020-06-01
Box Select Column

JamesIves:
Box select column changes for mobile view.


components/molecules/BoxSelect/BoxSelect.jsx
```diff
@@ -75,7 +75,7 @@ class BoxSelect extends Component {
         {options.map((item) => (
           <div
             key={item.value}
-            className="uic--col-6 uic--col-sm-4"
+            className="uic--col-sm-6 uic--col-md-4 uic--col-12"
```
RohitRajendran, line 78:
```suggestion
            className="uic--col-12 uic--col-sm-6 uic--col-md-4"
```
Nit picky but just rearranging it from small to bigger breakpoint

PR #1279, opened 2020-06-02
[APP-5100] TableTile

jamaalalleyne-c1:
Addressed vertical alignment of text for table tile stories

Regarding the TableTile stories ensure that the text displayed in each cell is vertically aligned.

- [APP-5100](https://unitedincome.atlassian.net/browse/app-5100)


components/molecules/TableTile/TableTile.scss
```diff
@@ -20,7 +20,7 @@
     border-color: $light-gray;
     border-style: solid;
     border-width: 1px 0 0;
-    padding: 1.168rem 1.8rem 1.772rem;
+    padding: 0 1.8rem 0;
```
RohitRajendran, line 23:
```suggestion
    padding: 0 1.8rem;
```
This could be simplified a little bit

PR #1290, opened 2020-06-08
Cypress data cy updates

alexgumus:
This PR carries all data-cy changes for the **component** side. With this update, these specific attributes will no longer rely on language based selectors from the client side.

All related PRs:

- Client
[data-cy attribute client changes](https://github.com/UnitedIncome/client/pull/4060)
- Cypress
[data-cy_attributes_changes](https://github.com/UnitedIncome/cypress-integration-testing/pull/26)


components/atoms/Slider/Slider.jsx
```diff
@@ -52,7 +52,7 @@ const Slider = ({
   };
 
   return (
-    <div className={containerClasses}>
+    <div className={containerClasses} data-cy="slider">
```
bryanjcof, line 55: It looks like there is a required `name` prop passed into `Slider`, does it make sense to use `name` here or `slider-${name}`?

components/molecules/RadioButtons/RadioButtons.jsx
```diff
@@ -14,6 +14,7 @@ const RadioButtons = ({
   value,
   onChange,
   className,
+  //datacy,
```
bryanjcof, line 17: If this isn't needed please remove it rather than leave it commented out.

components/molecules/RadioButtons/RadioButtons.jsx
```diff
@@ -68,6 +69,7 @@ const RadioButtons = ({
                 <span className={labelWrapper}>
                   <input
                     name={name}
+                    //data-cy={datacy || ''}
```
bryanjcof, line 72: If this isn't needed please remove it rather than leave it commented out.

components/organisms-simple/CompletionScreen/CompletionScreen.jsx
```diff
@@ -23,11 +23,14 @@ const CompletionScreen = ({
...
   return (
-    <div className={containerClass}>
+    <div className={containerClass} data-cy={title}>
       {feature && <div className="uic--feature">{feature}</div>}
       <h2>{title}</h2>
       <p className="uic--text-center">{description}</p>
-      <div className="uic--actions uic--d-flex uic--flex-column uic--align-items-center">
+      <div
+        className="uic--actions uic--d-flex uic--flex-column uic--align-items-center"
+        data-cy="continue"
```
bryanjcof, line 32: It looks like this could hold a more generic set of things than only "continue", should this be `actions` instead?

components/utilities/Confirm/Confirm.jsx
```diff
@@ -17,10 +17,10 @@ const Confirm = ({name, title, description, okLabel, cancelLabel}) => {
         <h2>{title}</h2>
         {description && <p className="uic--font-italic">{description}</p>}
         <div className="uic--modal-bottom">
-          <Button variant="secondary" light onClick={resolve}>
+          <Button variant="secondary" light onClick={resolve} data-cy="yes">
```
bryanjcof, line 20: Do these need to be more specific to the type of element, `confirm-yes` instead of `yes` for example?

components/molecules/RadioButtons/RadioButtons.jsx
alexgumus: Sure, will remove it.

components/organisms-simple/CompletionScreen/CompletionScreen.jsx
alexgumus: Tried all possible values. All shows up as `data-cy="Object object"`, had to go with static value.

components/utilities/Confirm/Confirm.jsx
alexgumus: Tried to keep consistency between displayed button names and these attributes. For this selector the button shows as `yes`

But I can change it to `data-cy=confirm-yes` from `data-cy="yes"`as well

components/molecules/RadioButtons/RadioButtons.jsx
alexgumus: Will remove. Thanks for catching these.

components/atoms/Slider/Slider.jsx
alexgumus: Sure, I will change it to `data-cy={name}` instead.

components/organisms-simple/CompletionScreen/CompletionScreen.jsx
bryanjcof: In the case of a static value, does a value of "actions" make more sense? It may not, I don't know how this is used so "continue" may well be better!

components/utilities/Confirm/Confirm.jsx
bryanjcof: Oh I see, that makes sense. I was only thinking about having multiple `Confirm` components on a screen and then using `data-cy` as a selector doesn't give a unique enough reference. I suppose this is unlikely to happen with a thing called `Confirm` but because this is in a shared components lib I think it is a good consideration when adding these attributes.

PR #1302, opened 2020-06-15
[APP-5295] Move `className` prop to root element of `CompletionScreen`

bryanjcof:
### Description

When a `className` prop is provided to a component it is expected to be
used on the root element of a complex component. This change moves the
usage of the `className` prop to the root element of the DOM tree
created by the `CompletionScreen` component.

### Testing Instructions

These changes should have functional difference to the previous version.
Any regression tests should still pass.

Refs [APP-5295](https://unitedincome.atlassian.net/browse/APP-5295)


bryanjcof:
I am not sure why this caused project coverage to decrease.

components/organisms-simple/CompletionScreen/CompletionScreen.jsx
```diff
@@ -12,19 +12,24 @@ const CompletionScreen = ({
 }) => {
   const containerClass = classNames(
     {
-      'uic--mcg-completion-screen': true,
-      'uic--align-items-center': true,
-      'uic--d-flex': true,
-      'uic--flex-column': true,
-      'uic--justify-content-center': true,
+      'uic--mcg-completion-screen-container': true,
```
JamesIves, line 15: I'd suggest putting `uic--mcg-completion-screen` here and putting `uic--mc-completion-screen__inner` on the inner div. It's more readable that way.

components/organisms-simple/CompletionScreen/CompletionScreen.jsx
```diff
@@ -12,19 +12,24 @@ const CompletionScreen = ({
...
+    'uic--flex-column': true,
+    'uic--justify-content-center': true,
+    'uic--w-100': true,
+  });
 
   return (
-    <div className="uic--mcg-completion-screen-container uic--w-100 uic--d-flex">
-      <div className={containerClass} data-cy={title}>
+    <div className={containerClass}>
+      <div className={completionClass} data-cy={title}>
```
JamesIves, line 32: `innerClass` would probably make more sense as a variable name here.

PR #1307, opened 2020-06-23
InputCard Optional Masks

ortizlu:
InputCard can have text as well, so there is no need for requiring a `mask` property. This makes it optional to include one.


components/organisms-complex/InputCard/InputCard.tsx
```diff
@@ -103,9 +103,9 @@ const InputCard: FC<InputCardProps> = ({
   // Automatically prepend/append depending on type of mask if answers is not passed through
   const answer = answers
     ? answers
-    : currencyMasks.includes(config.mask)
+    : currencyMasks.includes(config.mask || '')
```
RohitRajendran, line 106: Do you actually need the `|| ''`? Won't it just be false since currentMasks won't include undefined anyways

ortizlu: I think the `includes` method's type wants a string

PR #1321, opened 2020-07-01
Extract an IconProps type

dylanbyars:
### Description
A lot of the icon components shared a collection of types. I extracted those shared types into a single type that the components could share. 

### Comments
`ArrowLineIcon`, `HouseIcon`, and `InfoIcon` buck the trend of conformity so I `Omit` a prop from those. I'm wondering if people have opinions about this. 

At first, I started thinking that the `IconProps` that I'd pulled out weren't universal _enough_ to warrant getting extracted. Upon further reflection, it feels like calling out the strangeness of the previously mentioned icon components by tying them to the shared `IconProps` type makes their novelty more obvious.

### DIFF
--- a/components/atoms/icons/AddIcon/AddIcon.tsx
+++ b/components/atoms/icons/AddIcon/AddIcon.tsx
@@ -1,19 +1,10 @@
 /** @module AddIcon */
 import React, {FC, memo} from 'react';
+import {IconProps} from '../icons';
 
-type AddIconProps = {
+type AddIconProps = IconProps & {
   /** Title for the svg. */
   title?: string;
-  /** The width of the icon with unit sizing (px, rem, etc). */
-  width?: string;
-  /** The height of the icon with unit sizing (px, rem, etc). */
-  height?: string;
-  /** The color of the icon. */
-  fill?: string;
-  /** Additional class names to apply to the container. */
-  className?: string;
-  /** Additional inline styles to apply to the container. */
-  style?: React.CSSProperties;
 };
 
 /** Displays the AddIcon component.*/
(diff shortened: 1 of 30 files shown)


components/atoms/icons/ArrowLineIcon/ArrowLineIcon.story.jsx
```diff
@@ -15,7 +15,6 @@ stories.addParameters({
 const defaultProps = () => ({
   fill: text('fill', '#000'),
   height: text('height', '44.5rem'),
-  width: text('width', '1.3rem'),
```
dylanbyars, line 18: The `ArrowLineIcon` component doesn't recognize a width prop so I removed it.

components/atoms/icons/AddIcon/AddIcon.tsx
```diff
@@ -1,19 +1,10 @@
 /** @module AddIcon */
 import React, {FC, memo} from 'react';
+import {IconProps} from '../icons';
 
-type AddIconProps = {
+type AddIconProps = IconProps & {
   /** Title for the svg. */
   title?: string;
```
RohitRajendran, line 7: Should we just make `title` universal for all icons?

dylanbyars: I love it. Will do

dylanbyars: I'm going to update the storybook knobs of the components to account for this new prop but that'll take some time so I'm going to close this branch down and resume during our next innovation sprint.

dylanbyars: I'll save that improvement for the next innovation sprint. This work seems mergeable (if y'all agree).

PR #1329, opened 2020-07-15
[APP-5370] Go Directly to a Step in McGonagall

maggie9595:
### Description
1. If the user tries to navigate to a step that they've already completed, go directly to that step.
2. If the user tries to navigate to a step that they've NOT already completed, go to the first incomplete step.

Also added a story with a completed flow that you can test by adding a step query param (e.g. http://localhost:9001/iframe.html?id=templates-mcgonagall--completed&step=lastYearsTaxableIncome)

### Screenshots
New story: 
image

[Image from Gyazo](https://gyazo.com/85a6df12eb057a90593a80bab30915a4)

- [JIRA Ticket](https://unitedincome.atlassian.net/browse/app-5370)


components/templates/McGonagall/McGonagall.tsx
```diff
@@ -313,7 +316,32 @@ class McGonagall<TContext = DefaultContext> extends Component<
       }
     } while (latest.value !== newCardHistory[0].name && !finalReached);
 
-    const activeCard = this.stateValue(latest);
+    let activeCard = this.stateValue(latest);
+
+    // If a query param exists for the step and the step has been completed, show the summary step with the requested step as the active card
+    const queryStringCard = queryString.parse(queryParams).step as string;
```
dylanbyars, line 322: Is there a way to get this value without casting it to string?

The `activeCard` getter includes a check that guarantees the value parsed from the query string is a string. Maybe split that functionality out so both places can use it.

maggie9595: Good call, updated in 6fe24bf!

